PR #4, opened 2017-01-06
MAID-1922 : Access Containers FFI

gnunicorn:
- [x] setup temporary testing environment through FFI
- [x] Refresh Access Containers on register
- [x] Refresh MData on specific Container
- [x] Check container accessibility through FFI


gnunicorn:
All the Access Container tests currently fail (and are skipped) with Segfaults. Unclear where and why that happens.

PR #16, opened 2017-02-07
fix/reject the promise if the API args formatter throws an error

bochaco:
Even that this scenario may be due to programming errors, I think we should reject the promise to make it consistent with the case of input validation error coming from the FFI layer in which case the API would be rejecting the promise.

### DIFF
--- a/test/mutuable_data.js
+++ b/test/mutuable_data.js
@@ -7,32 +7,171 @@ const createAuthenticatedTestApp = h.createAuthenticatedTestApp;
 describe('Mutable Data', () => {
   const app = createAuthenticatedTestApp();
   const TAG_TYPE = 15639;
+  const TAG_TYPE_RESERVED = 10000;
+  const TAG_TYPE_INVALID = '_invalid_tag';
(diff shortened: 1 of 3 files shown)


test/mutuable_data.js
```diff
@@ -7,32 +7,171 @@ const createAuthenticatedTestApp = h.createAuthenticatedTestApp;
...
+  describe('Entries', () => {
+      it('get non-existing key', () => app.mutableData.newRandomPrivate(TAG_TYPE)
+          .then((md) => md.get('_non-existing-key'))
+          .then((value) => {
+            console.log("value:", value);
+            should(value).not.be.undefined();
+          })
+      );
+
+      it('insert and get a single entry', () => {
```
gnunicorn: please don't submit empty tests, they will be interpreted as "passing" which is misleading. Please instead throw an exception (like `throw new Error('Not Implemented')`) and for them to not be blocking your merges please mark them as to be skipped by using `it.skip(`.

bochaco: Ok, thanks, I will do that as well as the rebase.

PR #20, opened 2017-02-09
API Examples

shankar2105:
Integrated create public name and nfs read/write examples.


examples/nfs_file.js
```diff
@@ -0,0 +1,126 @@
...
+
+const readline = require('readline');
+const crypto = require('crypto');
+const safeApp = require('../src/');
+const lib = require('../src/native/lib');
+
+
+const appInfo = {
+  'id': 'net.maidsafe.examples.nfsReadWrite',
+  'name': 'NodeJS example App - NFS read/write',
```
gnunicorn: this breaks on Linux because of https://github.com/maidsafe/system_uri/issues/9 .

But let's fix that post merge.

PR #27, opened 2017-02-14
Tests for mutableData permissions and forEach functions

bochaco:


src/native/_mutable.js
```diff
@@ -31,6 +31,39 @@ function bufferLastEntry() {
         .concat([str, str.length]);
 }
 
+function callBackLastEntry() {
+  let fn = arguments[arguments.length - 1];
+  if (typeof fn !== 'function') throw Error('A function parameter _must be_ provided')
+
+  let cb = ffi.Callback("void", ['pointer', t.u8Pointer, t.usize],
```
gnunicorn: you use this - among others -  for the for_each_value. But value has the callback signature `fn(*mut c_void, *const u8,  usize, u64),` - with the last one representing the version number. I think we need to wrap that again into the `{buf: Buffer, version: u64}` we've used other places.

src/native/_mutable.js
```diff
@@ -194,13 +227,13 @@ module.exports = {
...
-    mdata_entries_for_each: Promisified(null, []),
+    mdata_entries_for_each: Promisified(callBackKeyValueLastEntry, []),
     mdata_entries_free: Promisified(null, []),
     mdata_keys_len: Promisified(null, t.usize),
-    mdata_keys_for_each: Promisified(null, []),
+    mdata_keys_for_each: Promisified(callBackLastEntry, []),
     mdata_keys_free: Promisified(null, []),
     mdata_values_len: Promisified(null, t.usize),
-    mdata_values_for_each: Promisified(null, []),
+    mdata_values_for_each: Promisified(callBackLastEntry, []),
```
gnunicorn: we need to do something more with the callback to wrap here. see comment before.

test/mutuable_data.js
```diff
@@ -187,9 +192,14 @@ describe('Mutable Data', () => {
         })
     );
 
-    it.skip('forEach on list of values', () => {
-      throw new Error('Test Not Implemented');
-    });
+    it('forEach on list of values', () => app.mutableData.newRandomPublic(TAG_TYPE)
+      .then((m) => m.quickSetup(TEST_ENTRIES).then(() => m.getValues()))
+      .then((values) => values.forEach((value) => {
```
gnunicorn: this must expect value + version, too

test/mutuable_data.js
```diff
@@ -340,47 +350,63 @@ describe('Mutable Data', () => {
...
-            )))
-    );
-
     it.skip('get permissions set', () => app.mutableData.newRandomPublic(TAG_TYPE)
         .then((m) => m.quickSetup(TEST_ENTRIES)
           .then(() => m.getPermissions()
             .then((perm) => app.auth.getPubSignKey()
-              .then((pk) => perm.getPermissionSet(pk.ref))
+              .then((pk) => perm.getPermissionSet(pk))
+              // the above command is failing with ERR_INVALID_SIGN_KEY_HANDLE
```
gnunicorn: is this still failing?

src/native/_mutable.js
bochaco: Yes, you are right, I'll try to fix it today.

PR #31, opened 2017-02-15
Fixed native library copy script.

ethanpailes:
When I tried to build this on linux the build script failed when it tried to copy the dynamic library from the wrong directory. This commit fixes things for me, though if you think it does the wrong thing let me know.

As a side node it seems like building safe_core without `--features "use-mock-routing testing"` does not compile right now. I assume that this is because there is stuff from routing/crust that is not ready yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,7 @@
     "build-system-uri": "cd system_uri && cargo build && cd ../ && cp -f system_uri/target/debug/*system_uri* ./src/native",
     "build-safe-lib": "cd native/safe_app && cargo build --features \"use-mock-routing testing\" && cd ../../ && cp -f native/target/debug/deps/*safe_app* ./src/native",
     "build-libs": "npm run build-system-uri && npm run build-safe-lib",
-    
-    "prepublish": "npm run build-libs",
+    "postinstall": "npm run build-libs",
     "pretest": "npm run build-libs",
     "docs": "documentation build --config etc/documentation.yml --github true --output docs --format html src/**",
     "serve-docs": "documentation serve --config etc/documentation.yml --github true --output docs --format html src/**",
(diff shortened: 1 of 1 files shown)


ethanpailes:
It looks like there are some differences between the environment on Travis and the one on my computer. Any insight?

gnunicorn:
@ethanpailes which rust versions are you using? I am running on Linux here and the script works fine the way it is. 

And yes, without mocked-routing it isn't compiling on `dev` at the moment, as not all of that is implemented in routing yet.

ethanpailes:
I just changed the name of the top level compile-on-install hook. `prepublish` runs when you `npm install` the root package, but not when you actually install it as a dependency. In reading about the issue is seems like the node community had decided that `prepublish` should pretty much only be used for running transpilers that emit javascript artifacts which have to end up in the final tarball. I think `postinstall` should be triggered at all the right times.

ethanpailes:
@gnunicorn I am using rust 1.15.1. Is that the wrong version?

gnunicorn:
thanks @ethanpailes, however, we've put that intentionally into `prepublish` for exactly the reasons you are describing. Our plan is to - later - offer to install (via npm) without any rust environment whatsoever. Instead we'll put a downloader-script into the post-install-hook - see https://maidsafe.atlassian.net/browse/MAID-1965 - that would fetch the precompiled libs for you. prepublish would then only be used in the development environment.

gnunicorn:
> I am using rust 1.15.1. Is that the wrong version?

There is a known bug in 1.15 with workspaces (which we use), which ends up not correctly building the lib. Could you try with 1.14?

ethanpailes:
Give me a sec

ethanpailes:
Building with 1.14.0 seemed to work. Thanks for the help. I'll close this PR and open another for the postinstall thing (which really is a separate issue).

gnunicorn:
>  open another for the postinstall thing (which really is a separate issue).

As commented before. The approach you proposed isn't the one we are intending to go with. No need to send a PR.

PR #32, opened 2017-02-16
fix/nfs_fetch: resolve nfs fetch parsing

shankar2105:
Resolved NFS fetch metadata parsing issue.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -86,7 +86,7 @@ class SAFEApp extends EventEmitter {
     return this.mutableData.newPublic(address, consts.TAG_TYPE_DNS)
       .then((mdata) => mdata.get(serviceName)
         .then((value) => this.mutableData.newPublic(value.buf, consts.TAG_TYPE_WWW))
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -86,7 +86,7 @@ class SAFEApp extends EventEmitter {
     return this.mutableData.newPublic(address, consts.TAG_TYPE_DNS)
       .then((mdata) => mdata.get(serviceName)
         .then((value) => this.mutableData.newPublic(value.buf, consts.TAG_TYPE_WWW))
-        .then((service) => service.emulateAs('NFS').fetch(path)));
+        .then((service) => service.emulateAs('NFS').fetch(path[0] === '/' ? path.substr(1) : path))));
```
gnunicorn: This entire part has been rewritten in https://github.com/maidsafe/safe_app_nodejs/pull/30

shankar2105: Yes. If the file name is `index.html` and it search for `/index.html` which throw file not found

PR #33, opened 2017-02-16
prepublish -> postinstall

ethanpailes:
I changed the name of the top level compile-on-install hook. `prepublish` runs when you `npm install` the root package, but not when you install it as a dependency. In reading about the issue is seems like the node community had decided that `prepublish` should pretty much only be used for running transpilers that emit javascript artifacts which have to end up in the final tarball. I think `postinstall` should be triggered at all the right times.

This blog has some more info on why prepublish behaves the way that it does: https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1#.mk97sy8ru

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "build-safe-lib": "cd native/safe_app && cargo build --features \"use-mock-routing testing\" && cd ../../ && cp -f native/target/debug/*safe_app* ./src/native",
     "build-libs": "npm run build-system-uri && npm run build-safe-lib",
     
-    "prepublish": "npm run build-libs",
+    "postinstall": "npm run build-libs",
     "pretest": "npm run build-libs",
     "docs": "documentation build --config etc/documentation.yml --github true --output docs --format html src/**",
     "serve-docs": "documentation serve --config etc/documentation.yml --github true --output docs --format html src/**",


ethanpailes:
I see that this is not the intended behavior in the long run as outlined here: https://maidsafe.atlassian.net/browse/MAID-1965. Would you be willing to have the build script as the postinstall hook as a stopgap measure so that people can use the library now? If not I can just keep a fork with the workaround, but it would be nice to be able to depend on the actual project.

gnunicorn:
@ethanpailes the only supported way right now is to use a local git-checkout as dependency. [I am working also making that work with a remote git-url](https://github.com/gnunicorn/safe_app_nodejs/tree/npm-ignore-with-libs) but we won't support actual `npm`-style-installs until we do have the package to publish, at which point we want to have the downloader there.

I am wondering though, how are you using the package at the moment? My problem was that with remote urls, npm didn't pick up the locally installed libs - but it _did_ ran prepublish. If that is the problem you faced, could you try [my npm-ignore-libs](https://github.com/gnunicorn/safe_app_nodejs/tree/npm-ignore-with-libs) branch and confirm that is a solution that works for you, too?

gnunicorn:
@ethanpailes this is what we are planning to do: https://github.com/maidsafe/safe_app_nodejs/pull/40

gnunicorn:
with #47 merged, we now have our system in place to do it the way we actually intended. Thanks for the patch @ethanpailes but it is now done in a different way :) .

PR #34, opened 2017-02-16
Updated README to include information on the right rust version.

ethanpailes:
Hopefully this will help prevent other people from making the same mistake that I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ safe_app nodejs library.
 
 # Development
 
-To build and use this module locally you need latest stable [NodeJS](https://nodejs.org/en/) (& npm) and [Rust](https://www.rust-lang.org/en-US/) (we recommend [rustup](https://rustup.rs/)).
+To build and use this module locally you need latest stable [NodeJS](https://nodejs.org/en/) (& npm) and [Rust](https://www.rust-lang.org/en-US/) (we recommend [rustup](https://rustup.rs/)). Note that Rust-1.15.0 and Rust-1.15.1 have some known issues with the way that cargo workspaces are handled, which is a feature we make use of. For this reason please use Rust-1.14.0 until Rust-1.16.0 is released.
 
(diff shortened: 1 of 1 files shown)


frabrunelle:
Hi @ethanpailes, we just updated our "Rust Style" document to include a section of which Rust version we currently use: https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Style.md#rust-version

Could you please update your PR to include that link instead? 🙂

I'm thinking that the first sentence could be updated to something like this:

> To build and use this module locally you need latest stable [Node.js](https://nodejs.org/en/) (& npm) and Rust ([click here to see which version we currently use](https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Style.md#rust-version)).

or feel free to let me know if you have another idea.

ethanpailes:
@frabrunelle Sure!

README.md
```diff
@@ -10,7 +10,8 @@ safe_app nodejs library.
 
 # Development
 
-To build and use this module locally you need latest stable [NodeJS](https://nodejs.org/en/) (& npm) and [Rust](https://www.rust-lang.org/en-US/) (we recommend [rustup](https://rustup.rs/)).
+To build and use this module locally you need latest stable [NodeJS](https://nodejs.org/en/) (& npm) and [Rust](https://www.rust-lang.org/en-US/) (we recommend [rustup](https://rustup.rs/)). [Click here to see which version we currently use](https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Style.md#rust-version).
```
frabrunelle: Since it's a separate sentence, I think it would be clearer to say: "Click here to see which version of Rust we currently use"

ethanpailes: I just changed it to explicitly mention rust.

ethanpailes:
Does that work?

frabrunelle:
yes 🙂

ethanpailes:
dope

PR #35, opened 2017-02-16
fix/delete & set permissionsSet and its tests

bochaco:


test/mutuable_data.js
```diff
@@ -356,23 +356,33 @@ describe('Mutable Data', () => {
           .then(() => m.getPermissions()
             .then((perm) => app.auth.getPubSignKey()
               .then((pk) => perm.getPermissionSet(pk).should.be.fulfilled())
-              // the above command is failing with ERR_INVALID_SIGN_KEY_HANDLE
-              // due to an issue in safe_app: MAID-????
+              // this is fixed in https://github.com/maidsafe/safe_client_libs/pull/459
```
gnunicorn: this has been merged.

gnunicorn:
@bochaco , looks like the last failing test is because the `AuthGranted` contains an optional `BootstrapConfig`

```Rust
/// safeclientlibs:/safe_core/src/ipc/resp/ffi.rs:L27
/// ----------------
/// It represents the authentication response.
#[repr(C)]
#[derive(Clone)]
pub struct AuthGranted {
    /// The access keys.
    pub app_keys: AppKeys,
    /// Access container
    pub access_container: AccessContInfo,

    /// Crust's bootstrap config
    pub bootstrap_config_ptr: *mut u8,
    /// `bootstrap_config`'s length
    pub bootstrap_config_len: usize,
    /// Used by Rust memory allocator
    pub bootstrap_config_cap: usize,
}
``` 

Which our current test doesn't contain. I am okay with either: disabling the test here and send a second PR to fix it, or fix it within this PR.

Fixing it means updating our JS-Struct and getting a newly generated response from authenticator (ask @shankar2105 ) to put into the test.

bochaco:
I'll send a separate PR for that then, so we can have this merged today.

PR #44, opened 2017-03-03
[WIP] MAID-1986 Investigating fastcall

gnunicorn:
- [x] switch to fast call
 - [x] implement fast call for app/auth
 - [ ] implement immutableData
 - [ ] implement mutableData

----

Broken ATM. Use the immutableData test!


gnunicorn:
Do we want to pick this up again, now that https://github.com/cmake-js/fastcall/issues/15 is fixed?

PR #46, opened 2017-03-08
MAID-1990 feat/Downloader Add Downloading of Safe App and pthread for Win

gnunicorn:


package.json
```diff
@@ -56,7 +54,28 @@
     "system_uri": {
```
bochaco: As a cosmetic comment, perhaps we can put all the elements that are meant for the dependency downloader under a single key, e.g.
```
"deps_downloader": {
     "system_uri": { ... },
     "safe_app": { ... },
     "ENV": { ... }
}

File: package.json [hunk 2]
```diff
@@ -16,10 +16,8 @@
     "build-system-uri": "cd system_uri && cargo build && cd ../ && cp -f system_uri/target/debug/*system_uri* ./src/native",
     "build-safe-lib": "cd native/safe_app && cargo build --features \"use-mock-routing testing\" && cd ../../ && cp -f native/target/debug/*safe_app* ./src/native",
     "build-libs": "npm run build-system-uri && npm run build-safe-lib",
```
bochaco: Are we keeping this for development purposes?

package.json
gnunicorn: They are: https://github.com/maidsafe/safe_app_nodejs/pull/46/files/076925214fc60ac0826bbae2856fed459d5d7bf0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L55

package.json
gnunicorn: Yes. Though we might want to reconsider the submodule-part now. It is a pain to download the entire repo on clone ..

package.json
bochaco: Ohh! I'm sorry, it shows I'm new with the github review tool :D

package.json
bochaco: Yes, it takes long. For development it's probably enough to be able to direct the deps. downloader to grab the libraries from my local cloned repo of the safe_app/system_uri.

PR #50, opened 2017-03-17
MAID-1978 - fix/adding optional bootstrap info to AuthGranted struct

bochaco:


bochaco:
Yes the auth-test works now, I'll be sending a PR for that now.

PR #57, opened 2017-03-23
MAID-2028 - Fixing MutableData permissions forEach binding function

bochaco:


src/api/auth.js
```diff
@@ -25,6 +25,7 @@ const nativeH = require('../native/helpers');
 const types = require('../native/types');
 const h = require('../helpers');
 const inTesting = require('../consts').inTesting;
+const { SignKey, EncKey } = require('./misc');
```
gnunicorn: I bluntly assume, you didn't do more than moving them out into their own file.

bochaco: Exactly, just so I can return a SignKey ref from both Auth and MutableData functions.

PR #74, opened 2017-05-26
feat/MutableData: adapt Private MutableData binding function to latest safe_app

bochaco:
Latest safe_app expects the secret key and nonce for creating a private MutableData.


bochaco:
@krishnaIndia , I took care of your requested changes in relation to the FFIResult for app_registered and app_unregistered.

PR #101, opened 2017-07-19
Resolve web fetch issue

shankar2105:
Resolve webfetch API to handle ‘/‘ as path, redirect to `index.html`.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -118,8 +118,6 @@ class SAFEApp extends EventEmitter {
                 let newPath;
                 if (!path || !path.length) {
                   newPath = '/index.html';
-                } else if (path === '/') {
-                  newPath = 'index.html';
                 } else if (path[path.length - 1] === '/') {
                   newPath = `${path}index.html`;
                 } else if (path[0] === '/') {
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -118,6 +118,8 @@ class SAFEApp extends EventEmitter {
                 let newPath;
                 if (!path || !path.length) {
                   newPath = '/index.html';
+                } else if (path === '/') {
+                  newPath = 'index.html';
```
bochaco: I think this is redundant with the following condition below: `else if (path[path.length - 1] === '/')`,  and it actually breaks the case of fetching something like `app.webFetch('safe://mydomain/')` (we should add this tests case, I pasted it below). I'm not sure which case you are trying to cover that is not currently covered by existing conditions.
``` javascript
  it('empty subdirectory fallback', function test() {
    this.timeout(20000);
    const content = `hello world, on ${Math.round(Math.random() * 100000)}`;
    return createRandomDomain(content, '/index.html', '')
      .then((domain) => createAnonTestApp()
        .then((app) => app.webFetch(`safe://${domain}/`)
          .then((f) => app.immutableData.fetch(f.dataMapName))
          .then((i) => i.read())
          .then((co) => should(co.toString()).equal(content))
      ));
  });
```

shankar2105: `else if (path[path.length - 1] === '/') {` condition is for accessing subdirectory `index.html` files, where `else if (path === '/') {` is for accessing `index.html` on rootDir

bochaco: I got you, so we are trying to support the case when the `index.html` is stored at root without a leading `/`. In that case I think we still need to work for fetching `safe://mydomain` (without trailing `/`).

PR #104, opened 2017-07-25
MAID-2128: Update safe_app_nodejs API to latest safe_client_libs NFS changes

hunterlester:


README.md
```diff
@@ -22,6 +22,12 @@ The external libraries will automatically be downloaded when you run `npm instal
 
 If you are working on a development environment, you can run `NODE_ENV=dev npm install` instead in order to get the `safe_client` libraries which use the `MockVault` file rather than connecting to the SAFE Network.
 
+## Testing
+
+You may possibly be compiling your own [safe_app](https://github.com/maidsafe/safe_client_libs/tree/master/safe_app) library for testing purposes.
+
+If you are using safe_app_nodejs in development, `NODE_ENV=dev`, it will import extra ffi function for testing. In this case, when compiling `safe_app` in `safe_client_libs` be sure to to include `testing` in your build features, i.e. `cargo build --release --features "use-mock-routing testing"`
```
bochaco: This is a bit confusing since you either compile the libraries using the testing features, or you use the `NODE_ENV=env` when doing `npm i` in safe_app_nodejs.

hunterlester: I'll double check that.

So, if `NODE_ENV=env` is flagged will a library dep be downloaded that is compiled with mock routing and testing features enabled? 

In my case, I needed to compile an updated version of safe_app that wasn't available for download as a dependency.

src/api/emulations/nfs.js
```diff
@@ -1,12 +1,33 @@
...
 
 function isString(arg) {
   return typeof arg === 'string' || (arg.toString ? arg.toString() === '[object String]' : false);
 }
 
 /**
+* Create a new file
+* @returns {File}
+**/
+function newFile() {
```
bochaco: Shouldn't we have this function as a method of some class, `NFS` or `File` ?

hunterlester: I originally had it as a method of NFS, however, the linter was complaining that the method didn't use `this`

src/api/emulations/nfs.js
```diff
@@ -181,11 +196,88 @@ class NFS {
...
+    const fileRef = file.ref ? file.ref.ref() : new File(file).ref.ref();
+    return lib.file_open(this.mData.app.connection, fileRef, openMode);
+  }
+
+  /**
+  * Get file size
+  * @param {FileContextHandle} fileContextHandle
+  * @returns {Promise<Number>}
+  **/
+  size(fileContextHandle) {
```
bochaco: I think this function should be a method of the `File` class which internally keeps its `fileContextHandle`. We don't expose handles from the nodejs API but only JS objects.

src/api/emulations/nfs.js
```diff
@@ -181,11 +196,88 @@ class NFS {
...
+  }
+
+  /**
+  * Read file
+  * @param {FileContextHandle} fileContextHandle
+  * @param {Number} position
+  * @param {Number} len
+  * @returns {Promise<[Data, Size]>}
+  **/
+  read(fileContextHandle, position, len) {
```
bochaco: I think this function should be a method of the `File` class which internally keeps its `fileContextHandle`. We don't expose handles from the nodejs API but only JS objects.

src/api/emulations/nfs.js
```diff
@@ -181,11 +196,88 @@ class NFS {
...
+    return lib.file_read(this.mData.app.connection, fileContextHandle, position, len);
+  }
+
+  /**
+  * Write file
+  * @param {FileContextHandle} fileContextHandle
+  * @param {Buffer|String} content
+  * @returns {Promise}
+  **/
+  write(fileContextHandle, fileContent) {
```
bochaco: I think this function should be a method of the `File` class which internally keeps its `fileContextHandle`. We don't expose handles from the nodejs API but only JS objects.

src/api/emulations/nfs.js
```diff
@@ -181,11 +196,88 @@ class NFS {
...
+    return lib.file_write(this.mData.app.connection, fileContextHandle, fileContent);
+  }
+
+  /**
+  * Close file
+  * @param {FileContextHandle} fileContextHandle
+  * @returns {Promise<File>}
+  **/
+  close(fileContextHandle) {
```
bochaco: I think this function should be a method of the `File` class which internally keeps its `fileContextHandle`. We don't expose handles from the nodejs API but only JS objects.

src/api/emulations/nfs.js
```diff
@@ -181,11 +196,88 @@ class NFS {
...
+    return lib.file_close(this.mData.app.connection, fileContextHandle);
+  }
+
 }
 
-module.exports = NFS;
+module.exports = {
+  NFS,
+  newFile
```
bochaco: We shouldn't need to export this, this should be just an internal helper function, and maybe only needed by the `NFS.open` function.

README.md
bochaco: Correct, the `NODE_ENV` is only used to signal which library to download and which is already built with/without testing & mock capabilities. If you build the library manually yourself, it then doesn't matter what you set to `NODE_ENV` when doing `npm i` as you are then manually replacing the libs in `src/native`

src/api/emulations/nfs.js
bochaco: We shouldn't need this at all, but if we do since it's imposed by the safe_client_libs, I think we would only need it within the `NFS.open` function, right here: https://github.com/maidsafe/safe_app_nodejs/pull/104/files#diff-edb2d18591ec6ad75cf8f1c8f46ce0dbR235

src/api/emulations/nfs.js
```diff
@@ -168,8 +180,11 @@ class NFS {
   * @returns {Promise<File>} - the same file
   **/
   insert(fileName, file) {
-    return lib.file_insert(this.mData.app.connection, this.mData.ref, fileName, file.ref.ref())
-      .then(() => file);
+    const ffiFile = new File(file);
```
bochaco: It sounds to me we don't need create a new File.

hunterlester: `new File` is ensuring that the given file is in the FfiFile format in safe_core and is also a ref type to send it as a reference to the network. 

I'm glad you point this out though because something is missing. I need to check if the file is already a ref type.

PR #116, opened 2017-08-08
MAID-2213 handle lib load failure

shankar2105:
Updated to handle lib load failure


src/error_const.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  FAILED_TO_LOAD_LIB: 1
```
krishnaIndia: This can be a object instead of a number. 
```
{
 code: 1,
 msg: (e) => { `Failed to load native libraries: ${e}` }
}
```

PR #119, opened 2017-08-09
MAID-2120

hunterlester:
informative error is raised with malformed appInfo

Is this what you all had in mind? Let me know.


src/app.js
```diff
@@ -23,6 +25,15 @@ class SAFEApp extends EventEmitter {
   */
   constructor(appInfo, networkStateCallBack) {
     super();
+
+    const appInfoMustHaveProperties = ['id', 'name', 'vendor', 'scope'];
+    const hasCorrectProperties = appInfoMustHaveProperties.every(
```
krishnaIndia: Just validating has own property wont be enough. Must check whether the value is not empty.

Can add an function `isValid` to the AppInfo class. This function can validate the instance and return boolean. Would be better to validate it in the model itself than doing it in the constructor.
```
isValid() {
   const mandatoryFields = ['id', 'name', 'vendor', 'scope'];
   for (let i = 0; i < mandatoryFields.length; i++) {
     if (!(mandatoryFields[i] && mandatoryFields[i].trim())) {
        return false;
     }
   }
   return true;
}
```

src/app.js
```diff
@@ -78,6 +82,23 @@ class SAFEApp extends EventEmitter {
...
 
+  /*
+  * Validates appInfo and properly handles error
+  */
+  validateAppInfo() {
+    const appInfo = this._appInfo;
+    const appInfoMustHaveProperties = ['id', 'name', 'vendor'];
+    let bool = false;
+    const hasCorrectProperties = appInfoMustHaveProperties.every((prop) => {
+      bool = Object.prototype.hasOwnProperty.call(appInfo, prop) && appInfo[prop];
```
bochaco: Shouldn't we also check that it's not an empty string after trimming white spaces from it?

test/index.js
```diff
@@ -17,6 +17,7 @@ describe('Smoke testing', () => {
 
 describe('External API', () => {
   describe('fromAuthURI', () => {
+<<<<<<< HEAD
```
bochaco: You'll have to resolve these conflicts.

PR #121, opened 2017-08-14
Maid 2238 - persist log file path 

krishnaIndia:
Persist the log file path after initialisation.

Add options to enable/disable rust logs, to register scheme with OS and also to pass path for loading the native libraries.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -20,10 +20,13 @@ class SAFEApp extends EventEmitter {
   * @param {AppInfo} appInfo
   * @param {Function} [networkStateCallBack=null] optional callback function
   * to receive network state updates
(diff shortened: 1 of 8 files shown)


src/app.js
```diff
@@ -216,9 +226,9 @@ class SAFEApp extends EventEmitter {
   */
   /* eslint-disable class-methods-use-this */
   logPath(logFilename) {
-    let filename = logFilename;
+    const filename = logFilename;
     if (!logFilename) {
-      filename = SAFEApp.logFilename;
+      return this.logPath;
```
bochaco: Shouldn't this be `return this.logFilePath` instead?

krishnaIndia: Resolved

PR #123, opened 2017-08-16
**WIP** MAID-2045: miscalculated timestamps and empty ref address on last close

hunterlester:


src/native/_base.js
```diff
@@ -68,7 +68,9 @@ module.exports = {
     },
     fromSafeLibTime: (sec, nsec_part) => {
       let d = new Date();
-      d.setTime(sec + (nsec_part / 1000));
+      // QUESTION: This doesn't fix any current bugs but shouldn't/
+      // sec be converted to milliseconds?
```
bochaco: I think they are milliseconds already, perhaps we can rename the variables to make it more clear to `sec_part` or something like that.

PR #126, opened 2017-08-17
MAID-2250 Update API to support latest system_uri

shankar2105:
Updated system_uri API to support latest binary.


src/native/_system_uri.js
```diff
@@ -11,24 +12,38 @@ let isSysUriLibLoadErr = null;
...
-      'string', //schemes
+      'string', //schemes,
+      'pointer', // userdata
+      'pointer'
     ] ],
   });
 };
 
+function _handleError() {
+  return FFI.Callback("void", [t.VoidPtr, t.FfiResult],
```
krishnaIndia: This will throw error out of the call stack. instead return promises as we do for all other APIs

PR #157, opened 2017-10-05
MAID-2389 feat/Add an option to allow registration of additional custom protocols in system URI scheme

bochaco:


src/api/auth.js
```diff
@@ -76,12 +73,21 @@ class AuthInterface {
...
+    const opts = this.app.options;
+    let scheme;
+    if (opts.registerScheme) {
+      scheme = `safe-${urlsafeBase64(appInfo.id).toLowerCase()}`;
+    }
+    if (opts.joinSchemes && opts.joinSchemes.length > 0) {
+      scheme = scheme ? [scheme].concat(opts.joinSchemes) : opts.joinSchemes;
+    }
+    if (scheme) {
+      lib.registerUriScheme({ bundle: appInfo.id,
```
joshuef: @bochaco I think we'll need the `dev` exclusion here for OSX to pass in the electron bundle identifier.

bochaco: Thanks @joshuef !
As agreed, I'll add a change to support a `bundle` field in the `appInfo`, and it fall backs to `appInfo.id` if not provided.

PR #159, opened 2017-10-17
MAID-2337 feat/Expose a set of constants which can be used by client apps

bochaco:


test/mutable.js
```diff
@@ -1,7 +1,7 @@
 const crypto = require('crypto');
 const should = require('should');
 const h = require('./helpers');
-const consts = require('../src/consts');
+const consts = require('../src/consts').pubConsts;
```
krishnaIndia: Can be refactored as `const {pubConsts: consts} = require('../src/consts')`

krishnaIndia: We can bring the import refactoring as a separate task

bochaco: I'll include those changes in this PR

bochaco:
@krishnaIndia , the changes in the example are already part of this PR: https://github.com/maidsafe/safe_app_nodejs/pull/159/files#diff-1553b81646f62e5efcd69a7f8aeab08aL33

src/api/emulations/nfs.js
```diff
@@ -1,7 +1,7 @@
 const lib = require('../../native/lib');
 const t = require('../../native/types');
 const nativeH = require('../../native/helpers');
-const consts = require('../../consts');
+const consts = require('../../consts').pubConsts;
```
joshuef: Let's import `consts` as `CONSTANTS` for consistency

src/api/mutable.js
```diff
@@ -3,7 +3,7 @@ const lib = require('../native/lib');
 const t = require('../native/types');
 const emulations = require('./emulations');
 const { SignKey } = require('./crypto');
-const CONST = require('../consts');
+const CONST = require('../consts').pubConsts;
```
joshuef: similarly here, `CONSTANTS` I think is clearest.

PR #161, opened 2017-10-23
refactor/Split the internal Promisified function which cover more specific cases separately

bochaco:


src/native/_base.js
```diff
@@ -74,7 +74,56 @@ module.exports = {
       d.setTime((sec * 1000) + (nsec_part / 1000000));
       return d;
     },
-    Promisified: function(formatter, rTypes, after, isLastArgCb) {
+    Promisified: function(formatter, rTypes, after) {
+      // create internal function that will be
+      // invoked ontop of the direct binding
+      // mixing a callback into the arguments
+      // and returning a promise
+      return (lib, fn) => (function() {
```
krishnaIndia: We can use `...varArgs` feature of es6 than using `arguments`.

src/native/_base.js
```diff
@@ -74,7 +74,56 @@ module.exports = {
...
+        // the internal function that wraps the actual function call
+        // compile the callback-types-definiton
+        let args;
+        let types = ['pointer', FfiResult]; // we always have: user_context and FfiResult
+        if (Array.isArray(rTypes)) {
```
krishnaIndia: Can be refactored using the spread operator
types.push(Array.isArray(rTypes) ? [...rTypes] : [rTypes]);

src/native/_base.js
```diff
@@ -74,7 +74,56 @@ module.exports = {
...
+          types = types.concat(rTypes);
+        } else if (rTypes) {
+          types.push(rTypes);
+        }
+        return new Promise((resolve, reject) => {
+          // if there is a formatter, we are reformatting
+          // the incoming arguments first
+          try {
+            args = formatter ? formatter.apply(formatter, arguments): Array.prototype.slice.call(arguments);
```
krishnaIndia: spread operator can be used instead of apply and slice?

src/native/_base.js
```diff
@@ -74,7 +74,56 @@ module.exports = {
...
+
+          // append user-context and callbacks to the arguments
+          args.push(ref.NULL);
+          args.push(ffi.Callback("void", types,
+              function(uctx, err) {
+                // error found, errback with translated error
+                if (err.error_code !== 0) return reject(makeFfiError(err.error_code, err.error_description));
+
+                // take off the ctx and error
+                let res = Array.prototype.slice.call(arguments, 2)
```
krishnaIndia: May be `const [ctxt, error, ...res] = varArgs` instead of slice?

krishnaIndia: Avoiding using `function` and stick to array functions for consistency?

src/native/_base.js
bochaco: I guess you meant `types.concat(...)` since you cannot push an array but concatenate it, but that still doesn't work since in some cases `rTypes` can be `null`.
I just pushed a new commit where I use the spread operator but just to replace the `concat` call.

src/native/_base.js
```diff
@@ -74,66 +74,108 @@ module.exports = {
...
+      // invoked on top of the direct binding
       // mixing a callback into the arguments
       // and returning a promise
-      return (lib, fn) => (function() {
+      return (lib, fn) => (function(...varArgs) {
         // the internal function that wraps the actual function call
         // compile the callback-types-definiton
         let args;
         let types = ['pointer', FfiResult]; // we always have: user_context and FfiResult
         if (Array.isArray(rTypes)) {
```
joshuef: Is this not needed in the `PromisifiedForEachCB`?

bochaco:
@joshuef , as pushed a new commit where I moved out some common code from the Promisified funcs, I think it looks a bit better. I kept the formatter step in them for now as I didn't like what I achieved when moving that out to a helper function which returns a promise.

PR #165, opened 2017-11-01
MAID-2413: travis deploy to npm

hunterlester:


.travis.yml
```diff
@@ -1,49 +1,27 @@
...
-  - npm run lint
-  - npm run test-coverage
-  - npm run publish-coverage
+- npm run lint
+- npm run test-coverage
+- npm run publish-coverage
+before_deploy:
+- rm -rf node_modules
+- unset NODE_ENV
+- npm install
```
joshuef: for consistency let's see if we can use `yarn` for this. Not sure if that's possible from the CI point of view (do they have `yarn` as a provider?). But worth a look. If not, then `npm` is fine for our purposes here.

krishnaIndia: We don't have to run npm install or yarn install for `safe-app`. It makes sense to install dev dependencies for `beaker-plugins-* ` since we have to use babel. The node_modules are ignored during publish. However in the script section we should be using `yarn` instead of npm. 

@joshuef, yarn is provided by [default](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn)

PR #167, opened 2017-11-06
MAID-2400: app_account_info

hunterlester:


src/native/_app.js
```diff
@@ -8,14 +8,40 @@ const { types } = require('./_auth');
...
+  },
   functions: {
+    app_account_info: [t.Void, [t.AppPtr, t.VoidPtr, 'pointer']],
     app_unregistered: [t.Void ,[t.u8Pointer, t.usize, t.VoidPtr, t.VoidPtr, 'pointer', 'pointer']],
     app_registered: [t.Void , ['string', ref.refType(AuthGranted), t.VoidPtr, t.VoidPtr, 'pointer', 'pointer']],
     app_reconnect: [t.Void, [t.AppPtr, t.VoidPtr, 'pointer']],
     app_free: [t.Void, [t.AppPtr]]
   },
   api: {
+    app_account_info: function(lib, fn) {
```
bochaco: You should be able to use the `Promisified` helper function from `_base.js`

test/app.js
```diff
@@ -106,4 +106,11 @@ describe('Smoke test', () => {
 
     should.throws(test);
   });
+
+  it.only('should return account information', async function test() { // eslint-disable-line prefer-arrow-callback
+    this.timeout(10000);
+    const app = createAuthenticatedTestApp();
+    const accInfo = await app.getAccountInfo();
+    console.log('account info: ', accInfo); // eslint-disable-line no-console
```
bochaco: We could as part of the test perform some write operations and verify they change in the `mutations_done`

src/native/_app.js
```diff
@@ -8,14 +8,27 @@ const { types } = require('./_auth');
...
 const consts = require('../consts');
 
+const AccountInfo = Struct({
+  mutations_done: t.u64,
+  mutations_available: t.u64
+});
+
 module.exports = {
+  types: {
+    AccountInfo
```
bochaco: I don't think we really need to export this

PR #169, opened 2017-11-08
MAID-2410: exposes network constants and functions on api

hunterlester:


hunterlester:
@bochaco Is this what you had in mind for this task?
I wasn't quite sure.

src/app.js
```diff
@@ -274,6 +261,44 @@ class SAFEApp extends EventEmitter {
     return this._appInfo;
   }
 
+  get CONSTANTS() { // eslint-disable-line class-methods-use-this
```
bochaco: We don't need to expose this here, we should just expose them from the global CONSTANTS object we are already exposing from here: https://github.com/maidsafe/safe_app_nodejs/blob/master/src/index.js#L4

src/app.js
```diff
@@ -243,20 +243,7 @@ class SAFEApp extends EventEmitter {
   * @param {String} state
   */
   set networkState(state) {
-    switch (state) {
```
bochaco: We need to have this conversion from a `number` to the textual representation somewhere, if it's not done from this function, then from the `get networkState()`: https://github.com/maidsafe/safe_app_nodejs/blob/master/src/app.js#L266

test/app.js
```diff
@@ -134,4 +135,30 @@ describe('Smoke test', () => {
...
   });
+
+  it('exposes network state constants', () => {
+    const app = createAuthenticatedTestApp();
+    should(app.CONSTANTS).have.properties(['INIT', 'CONNECTED', 'DISCONNECTED', 'UNKNOWN']);
+  });
+
+  it('returns boolean for network state: INIT', () => {
+    const app = createAuthenticatedTestApp();
+    should.exist(app.isNetStateInit);
```
bochaco: We should be checking if boolean value returned is correct.

src/app.js
bochaco: This is also affecting/changing current behaviour of the notifications, and that would break the current applications acting on the string returned in the notifications.

bochaco:
This was included in PR #171

PR #170, opened 2017-11-09
MAID-2416 test/Increase test coverage

hunterlester:


package.json
```diff
@@ -9,8 +9,8 @@
     "postinstall": "download_deps --package package.json",
     "docs": "documentation build --config etc/documentation.yml --github true --output docs --format html src/**",
     "serve-docs": "documentation serve --config etc/documentation.yml --github true --output docs --format html src/**",
-    "test": "mocha",
-    "test-coverage": "istanbul cover _mocha --report lcovonly -- -R spec",
+    "test": "mocha --recursive",
+    "test-coverage": "istanbul cover node_modules/mocha/bin/_mocha --report lcovonly -- -R spec --recursive test",
```
bochaco: We shouldn't need to specify the whole path to the `_mocha` command with `node_modules/mocha/bin/_mocha` as npm knows where to find it, isn't it working without the full path anymore?

src/api/immutable.js
```diff
@@ -94,8 +94,9 @@ class Writer extends helpers.NetworkObject {
   /**
   * Proxy to `close`.
   */
-  save() {
-    return this.close();
+  async save() {
```
bochaco: I would try to keep a consistency in the code to either use `Promises` or `async`, so we could plan for a refactoring to convert all the code to make use of `async` but until then I would say we stick to `Promises` to have a consistency in the code style.

test/app.js
```diff
@@ -8,21 +8,8 @@ const createAuthenticatedTestApp = h.createAuthenticatedTestApp;
...
-      });
-  });
-
-  it('should build some authentication uri', () => {
-    const app = createTestApp();
-    return app.auth.genAuthUri({ _public: ['Read'] })
-        .then((resp) => should(resp.uri).startWith('safe-auth:'));
-  });
+describe('Smoke test', function testContainer() {
+  this.timeout(15000);
```
bochaco: In PR #171  I got rid of `function` keywords to use arrow functions everywhere. Please let's stick to arrow functions. 
In this case that you need to set a timeout, you can set it after the function body, i.e. `describe('Smoke test', () => {...}).timeout(15000);`

test/app.js
```diff
@@ -150,4 +123,13 @@ describe('Smoke test', () => {
     await app.auth.loginFromURI(h.authUris.unregisteredUri);
     should(app.getAccountInfo()).be.rejected();
   });
+
+  it('returns safe_client_libs log path', async () => {
+    const app = createAuthenticatedTestApp();
+    app.logPath().should.be.fulfilled();
```
bochaco: In cases like this one, you need to return the `Promise` you get from `app.logPath()` otherwise the test won't fail if the `Promise` is rejected, you would only get an `unhandled promise` error but the test would still be passing. Please make sure you return `Promises` in all tests.

src/api/immutable.js
bochaco: Moreover, we can get rid of this function as it's not providing anything but just a name proxy, also I'm not sure if we want to have a function which by default stores data unencrypted, this could be error prone for the end users.

PR #171, opened 2017-11-10
MAID-2410 feat/Expose functions to verify current network connection state

bochaco:
Exposing some functions to verify current state of network connection. Also, replace all `functions` for arrow functions.

### DIFF
--- a/examples/standalone/scripts/idata_asymmetric.js
+++ b/examples/standalone/scripts/idata_asymmetric.js
@@ -1,8 +1,9 @@
 const safeApp = require('@maidsafe/safe-node-app');
+const { waitUntil } = require('wait');
 
 let EXIT_CONDITION = false;
 
-let run = async () => {
+const run = async () => {
 	const APP = {
         info: {
             id: 'net.safe.imd.demo.app',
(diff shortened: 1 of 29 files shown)


examples/standalone/scripts/idata_asymmetric.js
```diff
@@ -75,7 +75,7 @@ let run = async () => {
 
 run()
 
-function wait () {
```
joshuef: Is it possible to put the `wait` function in a helper/util file if we're using it all over? (can we pass in `EXIT_CONDITION`)?

test/index.js
```diff
@@ -17,6 +17,23 @@ describe('Smoke testing', () => {
   it('confirms the CONSTANTS are exposed', () => {
     should(CONSTANTS).be.Object();
   });
+
+  it('confirms the full list of contants', () => {
```
joshuef: `contants` => `con<s>tants`

examples/standalone/scripts/idata_asymmetric.js
bochaco: Yes we can do that, let me send a commit with such an utility function. The only comment is that if you grab one of these files to run them isolated you'll need to make sure to also include/copy that dependency file.

bochaco:
@joshuef , I pushed a commit based on your review. I'm using `wait` package from npm which provides the utility you suggested.

PR #172, opened 2017-11-10
MAID-2409 feat/API: Expose a function to read containers access permissions from an auth URI

bochaco:
This includes a single change needed to upgrade to `system_uri` lib v0.3.0.


src/api/auth.js
```diff
@@ -244,24 +244,51 @@ class AuthInterface {
...
 
+  /**
+  * Read granted containers permissions from an auth URI
+  * without the need to connect to the network.
+  * @arg {String} uri the IPC response string given
+  *
+  * @return {Promise<Array>}
+  */
+  readGrantedPermissions(uri) {
+    const sanitisedUri = removeSafeProcol(uri);
```
joshuef: `removeSafeProcol` -> `removeSafeProtocol`

src/native/_auth.js
```diff
@@ -132,12 +132,115 @@ const AuthGranted = Struct({
...
+    });
+  });
+
+  return new AccessContainerEntry({
+    containers: contInfoArray.buffer,
+    containers_len: contInfoArray.length
+  });
+}
+
+const makeAuthGranted = (authGrantedObj) => {
```
joshuef: is this 'Granting Authentication?' 

To me `AuthGranted` sounds like a boolean as opposed to a class. (not sure what is correct here, it's not that clear that this is doing for me)

src/native/_auth.js
```diff
@@ -132,12 +132,115 @@ const AuthGranted = Struct({
...
+    containersList.push({
+      name: currValue.name,
+      mdata_info: helpersForNative.makeMDataInfoObj(currValue.mdata_info),
+      permissions: helpersForNative.readPermsSet(currValue.permissions),
+    });
+  }
+  return containersList;
+}
+
+const readAuthGrantedPtr = (authGrantedPtr) => {
```
joshuef: `readAuthGrantedPtr`, what's 'Ptr'?

src/native/_auth.js
bochaco: `AuthGanted` is the name of the Node-FFI struct as defined in safe_app lib, this function just constructs the corresponding FFI-struct from a js object. I'm renaming it to `makeAuthGrantedFfiStruct`.
We can consider some refactoring to make a better naming and differentiation of the functions that convert FFI structs to js objects and vice versa.

src/native/_auth.js
bochaco: When we receive the `AuthGranted` FFI struct, we actually receive a pointer/ref to it, so `ptr` stands for pointer.

test/auth.js
```diff
@@ -135,6 +139,40 @@ describe('auth interface', () => {
   }).timeout(20000);
 });
 
+describe('Get granted containers permissions from auth URI', () => {
+  it('invalid uri', async () => {
+    const appNoConnect = createTestApp();
+    should(appNoConnect.auth.readGrantedPermissions('safe-invalid-uri')).be.rejectedWith('Serialisation error');
+  });
+
+  it('uri with no auth granted information', async () => {
```
joshuef: @bochaco this test... should it be rejected?

I mean... it's succeeding, there's just no perms granted to be shown, right?

bochaco: I think I understand what you mean, I was thinking that an auth URI like this one which is for unregistered connection (no containers permissions information in it even that it can be decoded) could be triggering an error to make the difference with a URI which does contain containers permissions (regardless the number of permissions). E.g. you can have an auth URI where you didn't request permissions for any container, therefore the returned object will be empty, I just added a test for this case and it expects to succeed and return an empty object.

Also, the other two types of auth URI's, i.e. for share MD and containers permissions, don't contain any information, they just have a request id and the containers permissions granted are just updated on the network, so for those cases the app will need to use the function `getContainersPermissions` after it's connected.

joshuef: Aha, yeh, that's what I was meaning.

If a non-authed app uri can never have permissions / would never be correct, then I'd say we change the error message to reflect this.

So we can have 1) authed app uri with no perms: return empty object.
2) unauthed app uri, which cannot have perms: throw error detailing that fact.

(maybe `This uri is not for an authenticated app` or something? so you can see your mistake)

PR #181, opened 2017-11-30
fix/fetch index.html with webFetch

hunterlester:


krishnaIndia:
The PR works only `index.html`. Other files are not loaded

PR #196, opened 2018-01-09
chore/update to Node ^8.0.0

hunterlester:


.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
 - linux
 language: node_js
 node_js:
-- '7.10.0'
+- '8'
```
bochaco: I think we should recommend ^8.0.0 but we need to strictly test with 8.0.0

PR #201, opened 2018-01-22
MAID-2130 feat/partial-content: Add option for partial content 

joshuef:
in webfetch requests.

Enables `webfetch` to accept an object with `url` and `range` set (for example; as well as a simple string), range then being used to retrieve partial content from a `File`

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -127,28 +127,29 @@ class SAFEApp extends EventEmitter {
   }
 
   /**
+  * @typedef {Object} WebFetchOptions
+  * holds additional options for the `webFetch` function.
+  * @param {Object} range range of bytes to be retrieved.
+  * The `start` attribute is expected to be the start offset, while the
+  * `end` attribute of the `range` object the end position (non inclusive)
+  * to be retrieved, e.g. with `range: { start: 1, end: 3 }` only the 2nd
+  * and 3rd bytes of data will be retrieved.
+  * If `end` is not specified, the bytes retrived will be from the `start` offset
+  * untill the end of the file.
+  * The ranges values are also used to populate the `Content-Range` and
+  * `Content-Length` headers in the response.
+  */
+
+  /**
   * Helper to lookup a given `safe://`-url in accordance with the
   * convention and find the requested object.
   *
-
   * @param {String} url the url you want to fetch
-  * @param {Object} options options Object
-  *                       range: { object } override header range with bytes from request.
+  * @param {WebFetchOptions} [options=null] additional options
   * @returns {Promise<Object>} the object with body of content and headers
   */
   async webFetch(url, options) {
     if (!url) return Promise.reject(new Error('No URL provided.'));
-
-    let range;
-    let start;
-    let end;
-    let contentLength;
-
-    if (options && options.range && typeof options.range.start !== 'undefined') {
-      range = options.range;
-      start = range.start;
-      end = range.end;
-    }
 
     const parsedUrl = parseUrl(url);
     if (!parsedUrl) return Promise.reject(new Error('Not a proper URL!'));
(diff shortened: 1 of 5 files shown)


hunterlester:
Tested and seen working nicely to fetch byte range. I see @bochaco's comments in Slack about `webFetch` taking additional options parameter instead of `urlOrReqObject`, so I'll wait to approve.

src/app.js
```diff
@@ -130,11 +130,27 @@ class SAFEApp extends EventEmitter {
   * Helper to lookup a given `safe://`-url in accordance with the
   * convention and find the requested object.
   *
+
   * @param {String} url the url you want to fetch
+  * @param {Object} options options Object
+  *                       range: { array }, start/end of partial content range to retrieve
```
bochaco: I just sent a PR to enhance the documentation for this by defining an object named `WebFetchOptions`.

src/app.js
```diff
@@ -130,11 +130,26 @@ class SAFEApp extends EventEmitter {
   * Helper to lookup a given `safe://`-url in accordance with the
   * convention and find the requested object.
   *
+
   * @param {String} url the url you want to fetch
+  * @param {Object} options options Object
+  *                       range: { object } override header range with bytes from request.
```
bochaco: I would define a custom object for documenting the options, let's say `WebFetchOptions`.

src/app.js
```diff
@@ -222,15 +237,35 @@ class SAFEApp extends EventEmitter {
           file = await emulation.fetch(filePath);
         }
         const openedFile = await emulation.open(file, consts.pubConsts.NFS_FILE_MODE_READ);
-        const data = await openedFile.read(
-          consts.pubConsts.NFS_FILE_START, consts.pubConsts.NFS_FILE_END);
+        const fileSize = await openedFile.size() || '*';
+        let data;
+
+        // TODO: how do we handle multipart Reqs
+        if (options && options.range) {
```
bochaco: We can remove the conditionals here and define start & end on L152 for both cases and then here keep it as:
```
data = await openedFile.read(start, end);
```

src/app.js
```diff
@@ -222,15 +237,35 @@ class SAFEApp extends EventEmitter {
           file = await emulation.fetch(filePath);
         }
         const openedFile = await emulation.open(file, consts.pubConsts.NFS_FILE_MODE_READ);
-        const data = await openedFile.read(
-          consts.pubConsts.NFS_FILE_START, consts.pubConsts.NFS_FILE_END);
+        const fileSize = await openedFile.size() || '*';
+        let data;
+
+        // TODO: how do we handle multipart Reqs
```
bochaco: We can either implement it here and have the `range` object to be renamed to `ranges` and be an array of objects instead, or just don't implement such a support here but in Peruse/browser, the former would be more efficient though as we would perform the steps of finding the root container for the service only once.

bochaco: I sent a PR (https://github.com/joshuef/safe_app_nodejs/pull/2) against your branch @joshuef which backs up my comments plus some more tests I created. I found that safe_app lib is not validating the start offset so we'll need to raise that as an issue, for now I skipped the tests corresponding to those scenarios.
You will see that I made an adjustment to the values returned in the `Content-Range` header as per my understanding of it.

src/app.js
```diff
@@ -222,15 +238,44 @@ class SAFEApp extends EventEmitter {
...
           headers: {
             'Content-Type': mimeType
           },
           body: data
-        });
+        };
+
+        if (range) {
+          endByte = (end === fileSize) ? fileSize - 1 : end;
+          response.headers['Content-Range'] = `bytes ${start}-${endByte}/${fileFinalByteIndex}`;
```
bochaco: I think the value after the `/` is incorrect, it should be the length not the last byte index

test/browsing.js
```diff
@@ -229,6 +229,107 @@ describe('Browsing', () => {
...
+
+      return createRandomDomain(content, '/streaming.mp4')
+        .then((domain) => createAnonTestApp()
+          .then((app) => app.webFetch(`safe://${domain}/streaming.mp4`,
+                    { range: { start: startByte, end: startByte + bytesToRetrieve } })
+            .then((data) => {
+              should(data.body.toString()).equal('lo wor');
+              should(data.body.toString()).equal(
+                          content.substring(startByte, startByte + bytesToRetrieve));
+              should(data.headers['Content-Range']).equal(`bytes ${startByte}-${startByte + (bytesToRetrieve)}/${numberOfBytes}`);
```
bochaco: I think `numberOfBytes` here shouldn't be the last byte index but the total length

test/browsing.js
```diff
@@ -229,6 +229,107 @@ describe('Browsing', () => {
...
+              should(data.headers['Content-Length']).equal(content.length - startByte);
+            })
+        ));
+    }).timeout(20000);
+
+    it('range end beyond data length', () => {
+      const content = `hello world, on ${Math.round(Math.random() * 100000)}`;
+      return createRandomDomain(content, '/streaming.mp4')
+        .then((domain) => createAnonTestApp()
+          .then((app) => should(app.webFetch(`safe://${domain}/streaming.mp4`, { range: { start: 0, end: content.length + 1 } }))
```
bochaco: Perhaps we can pass the `end ` with `content.length` rather than `content.length + 1` to make sure we test the limit.

joshuef: You mean in the HTTP req?

PR #206, opened 2018-02-06
chore/safe-app: Adapt code to upgrade safe-app lib to v0.6.0

bochaco:
The safe-app lib changed the signature for the test functions, i.e. `test_create_app` and `test_create_app_with_access`, also since they are now asynchronous functions our tests needed to be adapted.


joshuef:
tests passing for me on osx. looks good.

PR #207, opened 2018-02-07
feat/deps: enable both mock and live lib downloads.

joshuef:


src/consts.js
```diff
@@ -9,6 +10,24 @@ const NET_STATE_INIT = -100;
...
+// Determine if process is forced to mock
+const allPassedArgs = process.argv;
+
+let hasMockFlag = false;
+
+if (allPassedArgs.includes('--mock')) {
+  hasMockFlag = true;
+}
+
+const env = hasMockFlag ? 'development' : process.env.NODE_ENV || 'production';
```
bochaco: `hasMockFlag` will always hold a boolean value.

joshuef: What do you mean here? 

It does yep. But then were using that either as a hard set or falling back on system ENV.... / production.

PR #208, opened 2018-02-09
chore/auth: MAID-2503 throw informative error for container request on nonexistent containers

hunterlester:
…ent containers

The current behavior is that `safe_client_libs` returns null pointer if [access_container_fetch](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/access_container.rs#L56) is called on an app with no containers.

In `safe_app_nodejs`, if this is the case at the time of calling `getContainersPermissions` then the error, `Uncaught Error: reinterpret: Cannot reinterpret from NULL pointer`, is returned but not handled.

In the web API under the same conditions, `getContainersPermissions` hangs and returns no informative error.

@bochaco This PR is a WIP because it will be an on-going exploration of where we need to handle errors and to also prompt discussion about whether we should be handling the error or if, for example, [access_container_fetch](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/access_container.rs#L56) should be calling a separate error callback instead of using `o_cb` for all responses.


hunterlester:
Should `access_container_fetch` return an error instead of a null pointer for an empty container set for a given app?

hunterlester:
@bochaco I've confirmed with safe-core that `access_container_fetch` returning a null pointer for an empty container set is appropriate. PR ready for review.

src/native/_auth.js
```diff
@@ -214,7 +215,13 @@ module.exports = {
     access_container_fetch: helpers.Promisified(null, [ref.refType(ContainerPermissions), t.usize], (args) => {
       const ptr = args[0];
       const len = args[1];
-      let arrPtr = ref.reinterpret(ptr, ContainerPermissions.size * len);
+      let arrPtr;
+      try {
+        arrPtr = ref.reinterpret(ptr, ContainerPermissions.size * len);
+      } catch (err) {
+        const noContainersErr = errConst.NO_CONTAINERS;
+        return makeFfiError(noContainersErr.code, noContainersErr.msg);
```
bochaco: I don't think we should be generating an error in this case, what I get is that the safe_app lib returns a `null` pointer if the app doesn't have permissions for any container, so we should be making sure to handle that scenario and and generating an empty list of containers.
So I think here instead of doing the try/catch you just need to check for `len > 0` before trying to read the array, and that should be enough, the rest is covered by the `for` loop below.

test/auth.js
```diff
@@ -18,6 +19,12 @@ describe('auth interface', () => {
         .then((resp) => should(resp.uri).startWith('safe-auth:'));
   });
 
+  it('should throw error if non-standard container is requested', () => {
+    const containersPermissions = { _app: ['Read', 'Insert', 'ManagePermissions'] };
+    const result = createAuthenticatedTestApp('_test_scope', containersPermissions, { own_container: true });
+    should(result).be.rejected();
```
bochaco: Let's make sure it's rejected with the expected error, i.e. use `be.rejectedWith`

test/auth.js
```diff
@@ -89,6 +96,12 @@ describe('Access Container', () => {
       });
     })));
 
+  it('throws error if no containers found', async () => {
+    const appWithNoContainers = await createAuthenticatedTestApp();
+    const noContainersErr = await appWithNoContainers.auth.getContainersPermissions();
+    should(noContainersErr.message).be.equal(errConst.NO_CONTAINERS.msg);
```
bochaco: As mentioned above, we should expect just an empty list of containers here

src/native/_auth.js
hunterlester: Right, this makes much more sense and makes things simpler. Thank you.

src/app.js
```diff
@@ -60,7 +59,7 @@ class SAFEApp extends EventEmitter {
     // in this additional search path.
     if (this.options.configPath) {
       lib.app_set_additional_search_path(this.options.configPath)
-        .catch((err) => { console.error('Faled to set additional config search path', err); });
+        .catch((err) => console.error('Failed to set additional config search path', err));
```
joshuef: Lets `const` this up.

src/app.js
```diff
@@ -198,7 +194,7 @@ class SAFEApp extends EventEmitter {
         const serviceInfo = await getServiceInfo(lookupName, serviceName);
         if (serviceInfo.buf.length === 0) {
           const error = new Error();
-          error.code = ERR_NO_SUCH_ENTRY;
+          error.code = errConst.ERR_NO_SUCH_ENTRY;
```
joshuef: I think we should be consistent with whats in the consts. And so make this a `.code` over there.

src/error_const.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  ERR_NO_SUCH_DATA: -103,
```
joshuef: Lets add full objects + msgs for these for consistency

src/error_const.js
```diff
@@ -8,12 +11,83 @@ module.exports = {
     code: 2,
     msg: `
     Malformed appInfo.
-    Make sure you conform to proper format and that id, name, and vendor are defined:
+    Please conform to proper format and be sure id, name, and vendor properties are defined:
```
joshuef: > id, name, and vendor

`"id", "name and "vendor"`,  perhaps?

src/native/_mutable.js
```diff
@@ -127,21 +127,18 @@ const translatePrivMDInput = (xorname, tag, secKey, nonce) => {
   let name = xorname;
   if (!Buffer.isBuffer(xorname)) {
     const b = new Buffer(xorname);
-    if (b.length != 32) throw Error("XOR Names _must be_ 32 bytes long.")
```
joshuef: Just for my own clarity, @hunterlester , we're removing these checks as it's handled by the underlying libs?

test/app.js
```diff
@@ -132,6 +131,12 @@ describe('Smoke test', () => {
     should(app.getAccountInfo()).be.rejected();
   });
 
+  it('should throw error if no auth URI is present during login', async () => {
```
joshuef: if we're not using `await`, we can drop `async` here.

test/auth.js
```diff
@@ -18,16 +19,62 @@ describe('auth interface', () => {
...
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+    src\\test_utils.rs:89,23 in safe_app::test_utils
+
+    Err(
+        Unexpected(
+                "'_app' not found in the access container"
+                    )
+                    )
+
+                    ', C:\\Users\\viv\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\unwrap-1.1.0\\src\\lib.rs:67:24
```
joshuef: this all seems _very_ specific.

to windows/whitespace/viv

Viv-Rajkumar: yeh that guy always has issues lol. Not even sure where this came from. Don't remember me submitting anything with hardcoded paths :|

+1 to generic asserts ofc.

src/native/_mutable.js
hunterlester: Yeah, I'm removing them because our `XOR_NAME` ffi data type is already doing the checking and returns a helpful error.

src/native/_auth.js
```diff
@@ -214,7 +213,8 @@ module.exports = {
     access_container_fetch: helpers.Promisified(null, [ref.refType(ContainerPermissions), t.usize], (args) => {
       const ptr = args[0];
       const len = args[1];
-      let arrPtr = ref.reinterpret(ptr, ContainerPermissions.size * len);
+      if(len === 0) return [];
```
bochaco: This is wrong, you are returning an array when the function normally returns an object.

src/native/_mutable.js
bochaco: I was looking at this particular case last week as well, and since we are using `ref-array` for the `XOR_NAME`, `SYM_SECRET_KEY`, etc., then if we don't perform the length validation here the validation is made by `ref-array` rather than safe_app lib, but I don't think the error returned is that helpful, this is the type of error we get: `buffer length must be at least 32, got X`. So I'm not fully sure if we should stop returning our own message for this particular cases since you would get an error from `ref-array` but you wouldn't know which of the three args (`xorname`, `secKey` or `nonce`) is the one with the incorrect length.

src/api/crypto.js
```diff
@@ -54,18 +55,6 @@ class PubEncKey extends h.NetworkObject {
...
 
-  /**
-  * Decrypt the given cipher text (buffer or string) using this public
-  * encryption key and the given secret key
-  *
-  * @arg {Buffer} cipher to decrypt
-  * @arg {SecEncKey} secretEncKey secret encryption key
-  * @returns {Promise<Buffer>} plain text
-  */
-  decrypt(cipher, secretEncKey) {
```
bochaco: Why are we removing this function?

src/api/crypto.js
```diff
@@ -112,20 +102,9 @@ class SecEncKey extends h.NetworkObject {
...
-
-  /**
-  * Encrypt the input (buffer or string) using this secret encryption key
-  * and the recipient's public key
-  *
-  * @param {Buffer} data to be encrypted
-  * @param {PubEncKey} recipientPubKey recipient's public encryption key
-  * @returns {Promise<Buffer>} cipher text
-  */
-  encrypt(data, recipientPubKey) {
```
bochaco: Why are we removing this function?

src/app.js
```diff
@@ -122,7 +137,7 @@ class SAFEApp extends EventEmitter {
     });
 
     if (!hasCorrectProperties) {
-      throw makeFfiError(errConst.MALFORMED_APP_INFO.code, errConst.MALFORMED_APP_INFO.msg);
+      throw new Error(errConst.MALFORMED_APP_INFO.msg);
```
bochaco: So we are not returning error codes from safe-app-nodejs layer but just error messages? if so, do we need the `code` attribute in the `errConst` objects?
If we are using error codes, we would preferably add them to the documentation.

src/helpers.js
```diff
@@ -64,7 +66,30 @@ const autoref = (obj) => {
...
+      const bool = Object.prototype.hasOwnProperty.call(perm, prop) && perm[prop];
+      if (!bool) {
+        badPerm = perm;
+        return false;
+      }
+      return true;
+    }));
+
+  if (!hasCorrectProperties) {
+    throw makeFfiError(
```
bochaco: I see an inconsistency here since we are using `makeFfiError` to return an error which contains a `code`, but in other cases we are not returning an error code, even that both validations are made by safe_app_nodejs, e.g. here: https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-c9616fc06535a5565aa65b19f33fa08aR181, or  here https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-bd9c9dcd314f2d7df52935b3a6a4d504R140.

Perhaps we do want to return error codes from safe_app_nodejs, and we can use the same helper function to do so (we may want to rename `makeFfiError` in that case though), and we'd need to add the list to our documentation.

src/native/_auth.js
```diff
@@ -143,7 +143,6 @@ const makeAppInfo = (appInfo) => {
 
 const translateXorName = (str) => {
   const b = new Buffer(str);
-  if (b.length != 32) throw Error("XOR Names _must be_ 32 bytes long.")
```
bochaco: Same comment about this in `translatePrivMDInput` function would be applicable there as the user wouldn't know which param is wrong from the permissions array...??

src/api/crypto.js
hunterlester: I was confused about this part of the API. Why do `PubEncKey` and `SecEncKey` each have an `encrypt` and `decrypt` function?

I don't see that reflected in web API:
- https://github.com/joshuef/peruse/blob/master/app/extensions/safe/api/crypto_public_encryption_key.js#L3
- https://github.com/joshuef/peruse/blob/master/app/extensions/safe/api/crypto_secret_encryption_key.js#L3

src/api/crypto.js
hunterlester: Same here: I was confused about this part of the API. Why do `PubEncKey` and `SecEncKey` each have an `encrypt` and `decrypt` function?

I don't see that reflected in web API:
- https://github.com/joshuef/peruse/blob/master/app/extensions/safe/api/crypto_public_encryption_key.js#L3
- https://github.com/joshuef/peruse/blob/master/app/extensions/safe/api/crypto_secret_encryption_key.js#L3

hunterlester: Isn't the `encrypt` method redundant in `SecEncKey`, since I can get a recipients public key from raw bytes and then `encrypt` using the `PubEncKey` method?

src/native/_auth.js
hunterlester: This error won't be necessary anymore as it will have been caught by [validateMDataPermissions](https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-c9616fc06535a5565aa65b19f33fa08aR182).

src/app.js
hunterlester: I don't find the error codes useful but I was continuing with what was already established for consistency. The case where error codes are useful is when we are expecting certain native errors in `webFetch` and can just check for the presence of code `-103`, for example.
I suppose a similar case may be useful to developers building on top of web API or safe_app_nodejs later on.

hunterlester:
@bochaco Before I get too far into documenting errors, checkout my example that I pushed with my latest commit to see if it makes sense.

src/native/_auth.js
bochaco: I don't see that `validateMDataPermissions` is in any possible execution flow of a call to this `translateXorName` function

src/app.js
bochaco: If we are not adding error codes now, then we sholdn't need the error code in the `src/error_const.js` objects, right?

test/auth.js
```diff
@@ -18,16 +19,51 @@ describe('auth interface', () => {
         .then((resp) => should(resp.uri).startWith('safe-auth:'));
   });
 
+  it('throws error if missing permissions object', async () => {
```
bochaco: I don't think this test is correct, we should allow to call this function without any permissions object.

test/index.js
```diff
@@ -42,7 +42,6 @@ describe('Smoke testing', () => {
 
   it('requires additional functions for testing, if in non-production', () => {
     const testingApi = api[api.length - 1];
-    should(LIB_CONSTANTS.inTesting).be.true; // eslint-disable-line
```
bochaco: Why are we removing this assertion?

hunterlester: I originally incorrectly failed to include the the parentheses for `true()`, so the assertion is actually doing nothing. However, when I do add the parentheses, it fails on CI tests because `LIB_CONSTANTS.inTesting` is not a boolean but an array of values which are not the same between my local build and the CI build.

bochaco: Yeah, it seems the `inTesting` function is incorrectly defined, the use of the `match` function is incorrect as it doesn't return a boolean, you can use something like ```new RegExp("dev|development|testing|test").test(str);```

test/auth.js
hunterlester: ahh right. thanks for catching this. Should still be returning auth URI.

test/mutable/index.js
```diff
@@ -73,6 +79,32 @@ describe('Mutable Data', () => {
             })
     );
 
+    it('throws error if custom public is created with name not equal to 32 bytes', async () => {
```
bochaco: This seems a duplicate of this test: https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-de3a919540b70d521ff29a18e30c207aR40

test/mutable/index.js
```diff
@@ -73,6 +79,32 @@ describe('Mutable Data', () => {
...
+      const name = 'non XOR name';
+      const test = () => app.mutableData.newPublic(name, TYPE_TAG);
+      should(test).throw(errConst.XOR_NAME.msg(32));
+    });
+
+    it('throws error if custom private is created with name not equal to 32 bytes', async () => {
```
bochaco: This seems duplicate of test: https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-de3a919540b70d521ff29a18e30c207aR45

src/api/auth.js
```diff
@@ -174,6 +177,13 @@ class AuthInterface {
   * ])
   */
   genShareMDataUri(permissions) {
+    if (!permissions) {
```
bochaco: I think this validation is redundant with the one done within `validateShareMDataPermissions`, so we should be able to remove it form here. Also, the validation underneath is probably also better to have it within the `validateShareMDataPermissions`.

test/auth.js
```diff
@@ -36,6 +72,52 @@ describe('auth interface', () => {
...
+    const test = () => app.auth.genShareMDataUri(perms);
+    should(test).throw('writeUInt64: no digits we found in input String');
+  });
+
+  it('should throw error for share MD request if name is not 32 byte buffer', async () => {
+    const app = h.createTestApp();
+    const mdName = 'not 32 byte buffer';
+    const perms = [{ type_tag: 15001, name: mdName, perms: ['Insert'] }];
+    const test = () => app.auth.genShareMDataUri(perms);
+    should(test).throw(`buffer length must be at least 32, got ${mdName.length}`);
```
bochaco: We had the code to return the proper error for this case: https://github.com/maidsafe/safe_app_nodejs/pull/208/files#diff-bdb8c45f3dc853fdcc4b43507d6ce996L146

PR #217, opened 2018-03-09
feat/package: Whitelist the files for the npm package generated rather than having a blacklist

bochaco:


joshuef:
looks good @bochaco . I'm in favour of this over a blacklist. Should be more robust 👍

bochaco:
Yes, I agree, we obviously cold have the issue of creating new files and not adding them to the whitelist but it's better to have a broken package/functionality than leaking unwanted files mistakenly.

PR #221, opened 2018-04-13
WIP: Enable force of devmode

joshuef:


src/consts.js
```diff
@@ -94,23 +116,24 @@ const pubConsts = {
   MD_PERMISSION_EMPTY: 0,
 };
 
-const LIB_FILENAME = {
+const getLibFilename = () => ({
   win32: path.join('./', libLocationModifier, 'safe_app.dll'),
```
joshuef: @bochaco, while this PR works, I think actually it's simpler if we remove this `libLocationModifier` and make this purely the filename as it was previously, and then force the `customLibPath` to include `prod` or `mock`.

That would mean some smaller rewrite of apps, but would be cleaner in the long run (as opposed to this 'opt-in' approach)....

Thoughts?

joshuef:
I'd opt to close this in favour of https://github.com/maidsafe/safe_app_nodejs/pull/222. Thoughts @bochaco, @hunterlester ?

src/consts.js
hunterlester: I definitely prefer PR #222

bochaco:
@joshuef , based on our talk, I'd prefer to keep something like this one so the user doesn't have to worry about the path of the native libs up front, i.e. unless the user really needs to provide a different path then it shoudl be much easier if the user can just set a flag saying I want mock/prod and nothing else. Additionally and optionally, the user could also be allowed to provide a path.

bochaco:
@joshuef I just merged this branch onto PR 230 to continue with this. Let's continue the discussions and review there.

PR #222, opened 2018-04-13
feat/libs: Remove assumption for lib location for custom paths.

joshuef:
Default to prod unless customPath specified


joshuef:
I guess this would be a major bump as it's breaking for current app setups, but IMO, is probably the way to go.

PR #223, opened 2018-04-27
MAID-2488 feat/webFetch: impl multipart range requests

hunterlester:
If a multipart range is desired, the range value may be an array instead of an object.
```
const range = [
        {
          start: 2,
          end: 4
        },
        {
          start: 7,
          end: 9,
        },
        {
          start: 13,
         end: 17,
        },
];
```

What is returned is still an object with `headers` and `body` expect the body is an array of objects with `headers` and `body` properties:
```
{
  "headers": {
       "Content-Type": "multipart/byteranges",
       "Content-Length": 200
   }.
   body: [
       {
           "headers": {
                "Content-Type: "video/mp4,
                "Content-Range":  bytes 4-15/200
            },
            body: <Buffer>
       },
        {
           "headers": {
                "Content-Type: "video/mp4,
                "Content-Range":  bytes 40-99/200
            },
            body: <Buffer>
       },
        {
           "headers": {
                "Content-Type: "video/mp4,
                "Content-Range":  bytes 150-175/200
            },
            body: <Buffer>
       },
   ]
}
```

This seemed most appropriate way to structure response based on https://stackoverflow.com/questions/41348421/how-to-properly-parse-multipart-byteranges-responses-using-typescript-javascript, https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests, and https://github.com/felixge/node-formidable/issues/390


src/app.js
```diff
@@ -168,9 +168,6 @@ class SAFEApp extends EventEmitter {
     if (!url) return Promise.reject(makeError(errConst.MISSING_URL.code, errConst.MISSING_URL.msg));
 
     const parsedUrl = parseUrl(url);
-    if (!parsedUrl) {
```
bochaco: Why this is not needed anymore?
Also, we need to update the documentation as it seems it now accepts either an object or an array.

src/app.js
```diff
@@ -260,32 +257,60 @@ class SAFEApp extends EventEmitter {
...
+                'Content-Range': `bytes ${partStart}-${partEndByte}/${fileSize}`
+              }
+            };
+          }));
+        } else {
+          // handles non-partial requests and also single partial content requests
+          data = await openedFile.read(start, lengthToRead);
+        }
+
+        const mimeType = multipart ? 'multipart/byteranges' : false
```
bochaco: Do we need the `false` in the boolean operation?
I would rather break this into and if/else as it's a bit confusing and error prone.

src/app.js
```diff
@@ -260,32 +257,60 @@ class SAFEApp extends EventEmitter {
...
+
+        if (options && options.range && rangeIsArray) {
+          // block handles multipart range requests
+          data = await Promise.all(range.map(async (part) => {
+            const partStart = part.start;
+            const partEnd = part.end || fileSize - 1;
+            const partLengthToRead = (partEnd - partStart) + 1; // account for 0 index
+            const byteSegment = await openedFile.read(partStart, partLengthToRead);
+            const partEndByte = (partEnd === fileSize - 1) ? fileSize - 1 : partEnd;
+            const mimeType = mime.getType(nodePath.extname(filePath))
```
bochaco: It seems you don't need this within the iteration

test/browsing.js
```diff
@@ -326,6 +333,99 @@ describe('Browsing', () => {
             .be.rejectedWith('Invalid range start value')
         ));
     }); // .timeout(20000);
+
+    it('fetches multipart ranges', async () => {
```
bochaco: We should be testing the headers as we do for single range tests.

bochaco:
Perhaps we can return the array with headers and bodies but with an attributed called `parts` rather than `body`, it seems that's closer to what happens in HTTP since the repose contains multiple parts, e.g.:
```
{
  "headers": {
       "Content-Type": "multipart/byteranges",
       "Content-Length": 200
   }.
   parts: [
       {
           "headers": {
                "Content-Type: "video/mp4,
                "Content-Range":  bytes 4-15/200
            },
            body: <Buffer>
       },
       ...
   ]
}
```

test/browsing.js
```diff
@@ -93,6 +93,14 @@ describe('Browsing', () => {
       ));
   }).timeout(20000);
 
+  it('fetches file with non-explicit mime type', async () => {
```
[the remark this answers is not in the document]
hunterlester: This test differs with the one on line 193 by creating a file without an extension, so a mimetype cannot be detected.

src/app.js
hunterlester: This line isn't needed anymore because `!url` is already handled on line 168. If there is an error parsing the url it will be handled by the the `parseUrl` function.

hunterlester: What is it that accepts an object or an array? The `webFetch` function for it's `url` parameter? I just see that we handle objects for `options` parameter.

src/app.js
hunterlester: The mimeType variable? That should be different from the parent scope mimeType.

hunterlester:
@bochaco Yeah, I wasn't sure about this because I can't find documentation about how to encode multipart byte ranges with JSON: https://developer.mozilla.org/en-US/docs/Web/API/Response

test/browsing.js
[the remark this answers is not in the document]
bochaco: We could then add the validation for the mime types to the tests

src/app.js
bochaco: But the value you are setting to `mimeType` is always the same as it doesn't depend on any of the values in each iteration, right?

src/app.js
```diff
@@ -260,32 +257,64 @@ class SAFEApp extends EventEmitter {
...
-          range = options.range;
-          start = range.start;
+        if (options && options.range) {
+          rangeIsArray = Array.isArray(options.range);
           fileSize = await openedFile.size();
-          end = range.end || fileSize - 1;
-
+          range = options.range;
+          multipart = range.length > 1;
+          start = options.range.start || consts.pubConsts.NFS_FILE_START;
```
bochaco: So are we making the `start` optional as well? if so  we'd need to add tests for that

src/app.js
hunterlester: Currently `Promise.reject(makeError(errConst.MISSING_URL.code, errConst.MISSING_URL.msg));` is returned if `!parseUrl`, but this condition can never be met because Node's [url.parse](https://nodejs.org/docs/latest-v8.x/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) function will throw an error if no URL string is passed to it and also the function doesn't consider any string to be an invalid URL, it will simply return an [URL](https://nodejs.org/docs/latest-v8.x/api/url.html#url_class_url) object.

PR #229, opened 2018-05-07
MAID-2622 refactor/naming: aligns naming conventions between libraries

hunterlester:


src/api/auth.js
```diff
@@ -79,15 +79,15 @@ class AuthInterface {
...
+    this.setupURI();
   }
 
   /**
   * @private
   * Generate the app's URI for the IPC protocol using the app's id
   * and register the URI scheme.
   */
-  setupUri() {
+  setupURI() {
```
bochaco: I wouldn't do this change mainly because in this particular case it would break the camelCase convention/style we follow for JS and we are not really aligning it with any of the native lib functions I think. Thus I would instead change the `fromAuthURI` function we have in `index.js`.

bochaco:
Also, as discussed, let's please review the API's that are being exposed by the other bindings so we can have all bindings to expose similar and coherent API's

PR #231, opened 2018-05-09
MAID-2603 feat/nfs: allows the input of user metadata

hunterlester:


src/api/emulations/nfs.js
```diff
@@ -83,6 +68,14 @@ class File {
     return this._ref.data_map_name;
   }
 
+  /**
+  *
+  * @returns {Buffer} user_metadata
+  */
+  get userMetadata() {
+    return this._ref.user_metadata_ptr.deref();
```
hunterlester: Currently unable to successfully deref this ponter.

test/nfs.js
```diff
@@ -200,12 +209,63 @@ describe('NFS emulation', () => {
     const size = await file.size();
     return should(size).be.equal(18);
   });
-//  describe('internal function', () => {
```
bochaco: Let's remove any commented code

src/api/emulations/nfs.js
```diff
@@ -277,20 +279,25 @@ class NFS {
   * @param {Number|CONSTANTS.NFS_FILE_MODE_OVERWRITE|
   *         CONSTANTS.NFS_FILE_MODE_APPEND|
   *         CONSTANTS.NFS_FILE_MODE_READ} [openMode=CONSTANTS.NFS_FILE_MODE_OVERWRITE]
+  * @param {String|Buffer} userMetadata
   * @returns {Promise<File>}
   */
-  open(file, openMode) {
+  open(file, openMode, userMetadata) {
```
bochaco: I'm not so sure about this, it doesn't sound correct to pass new metadata when you are are opening a file, in fact the only case when it's gonna be used is when we are creating a file with this function by passing a `null` file param, otherwise the metadata is ignored without the user noticing it.
Moreover, if I `create` a file, I think the metadata is not used either until I decide to `insert` it into the container (or `update` some other existing one) with some name.

test/nfs.js
```diff
@@ -200,12 +209,63 @@ describe('NFS emulation', () => {
...
+      .be.equal('text/javascript');
+  });
+
+  it('throws error if invalid user metadata type is passed while updating file', async () => {
+    const m = await app.mutableData.newRandomPublic(TYPE_TAG);
+    await m.quickSetup({});
+    const nfs = await m.emulateAs('nfs');
+    let file = await nfs.create('hello, SAFE world!', 'text/plain');
+    await nfs.insert('hello.txt', file);
+    should(file.ref.user_metadata_len).be.equal(10);
```
bochaco: We shouldn't do redundant steps in our tests, if this same verification (length of the metadata in this case) is made in previous tests we can skip it here, and go straight to the other type of verification we are aiming at, otherwise we have a set of tests which are bigger and bigger but don't actually provide more coverage.

src/api/emulations/nfs.js
hunterlester: Yes, that makes much more sense. Much cleaner as well.

test/nfs.js
hunterlester: It's being removed with this PR.

src/api/emulations/nfs.js
```diff
@@ -199,6 +192,7 @@ class NFS {
   /**
   * Helper function to create and save file to the network
   * @param {String|Buffer} content - file contents
+  * @param {String|Buffer} userMetadata - optional user metadata
```
bochaco: This param is not part of the function's signature

hunterlester: @bochaco Good catch. Thank you.

PR #232, opened 2018-05-11
MAID-2553 fix/systemuri: fix for passing empty URI string on Windows

hunterlester:
When an empty string, `''` (ASCII 00), is passed to our `system_uri` `openURI` on Windows and on OSX, file explorer is opened and promise is resolved, which differs on Linux where the result is a rejected promise. What ultimately gets passed to `ShellExecute`'s [lpFile](https://github.com/maidsafe/system_uri/blob/master/src/windows.rs#L37) property is `[0]`.

I built a local `system_uri` dll, where I passed `ptr::null()` to `lpFile` and even that produced the same file explorer behavior.
I also tried passing an empty `Vec` to `lpFile`, [seen here](https://github.com/hunterlester/system_uri/blob/test/src/windows.rs#L28), which causes a crash with very generic error name and code, respectively, `ELIFECYCLE` and `3221225477`. cc: @nbaksalyar 

The simple solution in this library is only call native `open_uri` if URI is provided, although the downside is that any other library utilizing `system_uri` on Windows would also have to handle this minor bug as well.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -69,6 +69,7 @@ describe('Smoke testing', () => {
   // when a URI with unregistered protocol is passed.
   // Until this is potentially resolved in `system_uri` lib, this test will exist
   // to highlight the difference.
+  // MAID-2553 was raised to solve this.
   it('openUri behaves differently, based on platform, when handling unregistered protocol', async () => {
     const app = await h.createTestApp();
     if (process.platform === 'win32') {
--- a/test/index.js
+++ b/test/index.js
@@ -65,29 +65,36 @@ describe('Smoke testing', () => {
     return should.exist(console.warn);
   });
 
-  it('system uri openUri function rejected when empty URL is passed', () => {
-    const app = h.createTestApp();
+  // TODO: openUri behaves differently on Windows than OSX or Linux,
+  // when a URI with unregistered protocol is passed.
+  // Until this is potentially resolved in `system_uri` lib, this test will exist
+  // to highlight the difference.
+  it('openUri behaves differently, based on platform, when handling unregistered protocol', async () => {
+    const app = await h.createTestApp();
+    if (process.platform === 'win32') {
+      // Why fulfilled? Because Windows opens a dialog for user to choose application
+      // to associate with unregistered protocol.
+      return should(app.auth.openUri('unregistered://resource')).be.fulfilled();
+    } else if (process.platform === 'linux') {
+      // Why fulfilled? Because Linux in Travis CI does not behave the same as on desktop env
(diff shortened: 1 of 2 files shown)


hunterlester:
@joshuef I may need your assistance testing `openUri` behavior on OSX.

test/index.js
```diff
@@ -65,15 +65,29 @@ describe('Smoke testing', () => {
...
+    const app = h.createTestApp();
+    return should(app.auth.openUri(' ')).be.rejected();
+  });
+
+  it('system uri openUri function rejected when null URL is passed', () => {
+    const app = h.createTestApp();
+    return should(app.auth.openUri(null)).be.rejected();
+  });
+
+  it('system uri openUri function rejected when undefined URL is passed', () => {
```
bochaco: I don't think there is a difference between this test and the one on L73

hunterlester:
Additionally, a URI like `unregistereduri://resource` doesn't return an expected error like `SE_ERR_NOASSOC`.
Instead, [ShellExecute](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762153(v=vs.85).aspx) opens dialog as follows:
unregisteredapp
Which returns an indication of success since that dialog was opened for the given URI.
This is inconsistent with default behavior on OSX and Linux, where an error is simply returned and promise is rejected.

@nbaksalyar I'm wondering if I may be able to get more control over the outcome of an open operation with winapi's [ShellExecuteEx](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762154(v=vs.85).aspx) function, which takes a [SHELLEXECUTEINFO](https://msdn.microsoft.com/en-us/library/windows/desktop/bb759784(v=vs.85).aspx) struct.
I'd like to test it but I'm not clear on how to import the struct in `system_uri`.

My goal is to simple have an error code and message returned instead of the dialog opening to choose an associated application.

test/index.js
```diff
@@ -65,15 +65,36 @@ describe('Smoke testing', () => {
...
-  // for the `openUri` function behaviour.
-  // On Linux the promise is rejected as expected, while on Windows
-  // the promise is resolved with an undefined value. We need to research
-  // where the issue exactly is, i.e. system_uri lib or safe_app_nodejs.
-  // MAID-2553 was raised to solve this.
-  it('system uri openUri function returns a promise', async () => {
+  // TODO: openUri behaves differently on Windows than OSX or Linux,
+  // when a URI with unregistered protocol is passed.
+  // Until this is potentially resolved in `system_uri` lib, this test will exist
+  // to highlight the difference.
```
bochaco: Let's keep the reference to MAID-2553 here.

PR #233, opened 2018-05-11
MAID-2414 feat/systemuri: add config option to disable system URI operations

hunterlester:
Do we also want to prevent `system_uri.(dll|so|dylib)` from downloading?

Currently working on this being an app initialization option, however, it is preferable to have a CLI flag, like `--mock`?


hunterlester:
Currently unable to get new tests to fail as intended.

PR #237, opened 2018-05-21
MAID-2564 feat/app: network state callback invoked when state changes from INIT to CONNECTED

hunterlester:


src/app.js
```diff
@@ -335,8 +335,7 @@ class SAFEApp extends EventEmitter {
   * Must be invoked when the client decides to connect back after the connection was lost.
   */
   reconnect() {
-    return lib.app_reconnect(this.connection)
-      .then(() => this._networkStateUpdated(null, consts.NET_STATE_CONNECTED));
+    return lib.app_reconnect(this);
```
bochaco: Isn't much simpler to just call `this._networkStateUpdated(null, consts.NET_STATE_CONNECTED);` from here, the `app_reconnect` function doesn't do anything else really.

test/app.js
```diff
@@ -129,16 +129,6 @@ describe('Smoke test', () => {
     should(app.reconnect()).be.fulfilled();
   });
 
-  it('throws error when reconnect operation called on unregistered app', async () => {
```
bochaco: Why are we removing this test? it seems the description/summary or the test itself is inaccurate as we are not calling reconnect here.

src/native/_app.js
```diff
@@ -74,19 +74,30 @@ module.exports = {
             const result = helpers.makeFfiResult(resultPtr);
             if (result.error_code !== 0) {
               reject(makeError(result.error_code, result.error_description));
-              return;
```
bochaco: Why are we removing this? the function will keep executing after calling the reject callback was invoked.

test/app.js
```diff
@@ -251,7 +241,59 @@ describe('Smoke test', () => {
...
+      .then(() => {
+        should(app.networkState).be.equal('Connected');
         app.auth.simulateNetworkDisconnect();
       });
   }).timeout(5000);
+
+  it('network state upon network reconnect event', (done) => {
+    let cbCount = 0;
+    const networkCb = () => {
+      if (cbCount === 1) {
```
bochaco: A more cleaner way is achieved with a switch statement, or with an array/map which contains the results expected for each `cbCount`. Perhaps worthy verifying the case for `cbCount === 0` as well.

src/app.js
hunterlester: Calling `_networkStateUpdated` here seemed out of place because because it's used [here](https://github.com/maidsafe/safe_app_nodejs/blob/master/src/native/_app.js#L48) and [here](https://github.com/maidsafe/safe_app_nodejs/blob/master/src/native/_app.js#L71) in `src/native/_app`.

PR #238, opened 2018-05-22
refactor/webFetch: split the main body of webFetch function into some internal helpers functions

bochaco:


src/web_fetch.js
```diff
@@ -62,59 +65,73 @@ async function webFetch(url, options) {
...
         serviceMd = await this.mutableData.fromSerial(serviceInfo.buf);
       } catch (e) {
         serviceMd = await this.mutableData.newPublic(serviceInfo.buf, consts.TAG_TYPE_WWW);
       }
-      const emulation = await serviceMd.emulateAs('NFS');
+
+      return serviceMd;
+    };
+
+    // Helper function to try different paths to find and
```
joshuef: @bochaco only thing I can think of here is maybe adding some tests for these helper functions. Otherwise all looking good

bochaco: Fully agreed, I think that should be the min benefit of splitting this function apart form readability. I'll work on it in the next few days.

src/web_fetch.js
```diff
@@ -158,39 +174,35 @@ async function webFetch(url, options) {
...
       if (range && multipart) {
-        response = {
-          headers: {
-            'Content-Type': mimeType,
-            'Content-Length': JSON.stringify(data).length
-          },
-          parts: data
-        };
+        response.headers['Content-Length'] = JSON.stringify(data).length;
+        response.parts = data;
```
hunterlester: Minor: If you place a log somewhere after this point you'll see both the `parts` and `body` properties in the `data` object.
```
{ headers:
   { 'Content-Type': 'multipart/byteranges',
     'Content-Length': 365 },
  body:
   [ { body: <Buffer@0x0000026677A49E10 6c 6c 6f>,
       headers: [Object] },
     { body: <Buffer@0x0000026677A49E18 6f 72 6c>,
       headers: [Object] },
     { body: <Buffer@0x0000026677A49E20 6c 6f 72 65 6d>,
       headers: [Object] } ],
  parts:
   [ { body: <Buffer@0x0000026677A49E10 6c 6c 6f>,
       headers: [Object] },
     { body: <Buffer@0x0000026677A49E18 6f 72 6c>,
       headers: [Object] },
     { body: <Buffer@0x0000026677A49E20 6c 6f 72 65 6d>,
       headers: [Object] } ] }
```
Should `body` be a null value or just `delete`d altogether?

bochaco: Nice catch @hunterlester , I think it should be `delete`ed, I will try to also add some tests which make sure the attributes returned are exactly those expected and no more for each case.

bochaco:
@joshuef , I created a [task](https://maidsafe.atlassian.net/browse/MAID-2683) so we can work on the additional tests and send a separate PR for them.

PR #242, opened 2018-05-31
MAID-2414 feat/mobile: prevents system_uri download based on deps_downloader

hunterlester:


bochaco:
@hunterlester , look at my suggestion in the PR #10 in deps_downloader, we can probably simplify it even further.
Also, I'm wondering if we could have an additional property supported by the deps_downloader which allow us to expand the rules of an artifact rather than override its attributes, e.g. if I set `override: false` in here https://github.com/maidsafe/safe_app_nodejs/pull/242/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L84, then the mock safe_app lib will be also installed in the mock location, by having this should  be able to get rid of the `install-dev-libs.js` script.

PR #243, opened 2018-05-31
MAID-1945 test/permissions: verifies that MAID-1945 is obsolete

hunterlester:


test/auth.js
```diff
@@ -39,19 +39,25 @@ describe('auth interface', () => {
...
 
-  it('should throw error if non-standard container is requested', () => {
+  it('should throw error if invalid container permissions format is passed', () => {
     const containersPermissions = { _app: ['Read', 'Insert', 'ManagePermissions'] };
     return should(createAuthenticatedTestApp('_test_scope', containersPermissions, { own_container: true }))
-      .be.rejected();
+      .be.rejectedWith("'_app' not found in the access container");
+  });
+
+  it('should throw error if non-standard container is requested', () => {
```
bochaco: I don't think this test adds anything to the previous one, we could keep this one and remove the previous perhaps, but both seem redundant to me.

test/auth.js
```diff
@@ -39,19 +39,25 @@ describe('auth interface', () => {
       .then((resp) => should(resp.uri).startWith('safe-auth:'));
   });
 
-  // TODO: issue MAID-2542 has been raised for safe_app lib as it's panicking
-  // instead of returning an error code and message.
-  // We'll need to adapt the test accordingly nce that's fixed.
   it('throws error if permissions object contains invalid permission', async () => {
     const app = await h.createTestApp();
     const test = () => app.auth.genAuthUri({ _public: ['Invalid'] });
     return should(test).throw('Invalid is not a valid permission');
```
bochaco: `genAuthUri` is still throwing an error instead of rejecting the promise?, perhaps it's now an issue on `genAuthUri` itself if MAID-2542 was properly fixed?

hunterlester: It looks like the [MAID-2542](https://maidsafe.atlassian.net/browse/MAID-2542) addressed invalid container names and not invalid permissions. See [PR](https://github.com/maidsafe/safe_client_libs/commit/b733c966cba10b3aad9993466474f31096febb6e).

This test seems fine to throw an error since we are [validating](https://github.com/maidsafe/safe_app_nodejs/blob/master/src/native/_base.js#L57) permissions in the library before sending it off to asynchronous FFI.

bochaco: So you mean it's ok because we always throw errors when doing validations?

hunterlester: Right, specifically for validating permissions, `safe_app_nodejs` will throw a synchronous error before calling the native operation.

PR #244, opened 2018-05-31
MAID-2664 test/helpers: simplifies test helper functions

hunterlester:


test/app.js
```diff
@@ -18,21 +18,18 @@ const errConst = require('../src/error_const');
...
-const createTestAppNoInit = h.createTestAppNoInit;
 const createAuthenticatedTestApp = h.createAuthenticatedTestApp;
-const createTestAppWithNetworkCB = h.createTestAppWithNetworkCB;
-const createTestAppWithOptions = h.createTestAppWithOptions;
 const { autoref } = require('../src/helpers');
 
 describe('Smoke test', () => {
   it('should return undefined value if log option is true, however app logging is not initialised', async () => {
-    const app = await createTestAppNoInit(null, { log: true });
+    const app = autoref(new App(h.appInfo, null, { log: true }));
```
bochaco: Can't we use `createTestApp` here too?

test/browsing.js
```diff
@@ -81,171 +80,143 @@ const removeFromMData = (md, key) => md.getEntries()
         )));
 
 describe('Browsing', () => {
+  let app;
+
+  before(async () => {
+    const containersPermissions = { _public: ['Read'], _publicNames: ['Read', 'Insert', 'ManagePermissions'] };
```
bochaco: Perhaps we can move this out as global const in the file?

test/browsing.js
```diff
@@ -81,171 +80,143 @@ const removeFromMData = (md, key) => md.getEntries()
...
+    app = await createAuthenticatedTestApp('_test_scope', containersPermissions);
+  });
+
   it('returns rejected promise if no url is provided', () => {
     const content = `hello world, on ${Math.round(Math.random() * 100000)}`;
-    return createRandomDomain(content, '', '')
-      .then(() => createAnonTestApp()
-        .then((app) => should(app.webFetch()).be.rejectedWith(errConst.MISSING_URL.msg)));
+    return createRandomDomain(content, '', '', app)
+      .then(() => should(app.webFetch()).be.rejectedWith(errConst.MISSING_URL.msg));
```
bochaco: Strictly speaking what you are doing here is not exactly the same, since you are using a single authenticated app for both publishing the site and for fetching it with `webFetch`, while the original code is using a separate un-authenticated app for webFetching the site, which is what happens when you are browsing sites.

Since the enhancement you are doing is really good as you moved the creation of the authenticated app to the `before` body, you can simply add a the creation of a second and un-authenticated app in the `before` as well, and it shouldn't impact the performance you gained with these changes.

test/cipher_opt.js
```diff
@@ -45,7 +47,14 @@ describe('CipherOpt', () => {
   });
 
   it('asymmetrically encrypts data to be written to immutable structure', async () => {
-    const differentApp = await h.createAltAuthTestApp();
+    let differentApp = autoref(new App({
```
bochaco: How about changing the `createTestApp` function to accept a `partialAppInfo` param instead of a `scope` param, which then merges it (`Object.assign(appInfo, partialAppInfo)`) with the default `appInfo`, by doing this we can call it from here with the modified appInfo we need here?

test/app.js
hunterlester: Here I need the app to be created without calling `app.init()`, as it is in [createTestApp](https://github.com/hunterlester/safe_app_nodejs/blob/c68e55a1d18525e7213b80643523508f50c289ee/test/helpers.js#L38).
This was the only test where I was using `createTestAppNoInit`.

PR #245, opened 2018-06-04
fix/docs: Update docs to reflect returned object from genAuth functions

joshuef:


src/native/_mutable.js
```diff
@@ -65,7 +65,7 @@ const MDataInfo = Struct({
   /// Name of the mutable data.
   name: t.XOR_NAME,
   /// Type tag of the mutable data.
-  type_tag: t.u64,
+  typeTag: t.u64,
```
bochaco: I'm not sure if it really matters but perhaps is worthy keeping the prop name here as `type_tag` as it's supposed to match the native struct.

src/native/_mutable.js
```diff
@@ -105,7 +105,7 @@ const makeMDataInfo = (mDataInfoObj) => {
   }
   return new MDataInfo({
     name: mDataInfoObj.name,
-    type_tag: mDataInfoObj.type_tag,
+    typeTag: mDataInfoObj.typeTag,
```
bochaco: I'm not sure if it really matters but perhaps is worthy keeping the prop name here as `type_tag` as it's supposed to match the native struct, and only call it `typeTag` on the json we expose only

src/native/_mutable.js
```diff
@@ -224,8 +224,8 @@ const makeMDataInfoObj = (mDataInfo) => {
   } catch (err) {
     throw makeError(errConst.XOR_NAME.code, errConst.XOR_NAME.msg(t.XOR_NAME.size));
   }
-  const type_tag = mDataInfo.type_tag;
-  if(!Number.isInteger(type_tag)) throw makeError(errConst.TYPE_TAG_NAN.code, errConst.TYPE_TAG_NAN.msg); 
+  const typeTag = mDataInfo.typeTag;
```
bochaco: This should probably be `const typeTag = mDataInfo.type_tag;`

src/native/_auth.js
```diff
@@ -43,7 +43,7 @@ const ContainerPermissionsArray = new ArrayType(ContainerPermissions);
 
 /// For use in `ShareMDataReq`. Represents a specific `MutableData` that is being shared.
 const ShareMData = Struct({
-  type_tag: t.u64,
+  typeTag: t.u64,
```
bochaco: I'm not sure if it really matters but perhaps is worthy keeping the prop name here as type_tag as it's supposed to match the native struct.

src/native/_auth.js
```diff
@@ -277,7 +277,7 @@ const makeShareMDataPermissions = (permissions) => {
   return new ShareMDataArray(permissions.map((perm) => {
     const permsArray = helpers.makePermissionSet(perm.perms);
     return ShareMData({
-      type_tag: perm.type_tag,
+      typeTag: perm.typeTag,
```
bochaco: I think here we also need to keep `typeTag: perm.typeTag,` as I think both structs (in and out) are used by ffi, i.e. they are internal not a json we expose.

src/native/_mutable.js
joshuef: I think this one stands, as the `mDataInfo` is a js construct from line 108 above,.

joshuef:
it's looking like the naming just needs to be consistent, and seems like the native FFI doesn't really mind what it's called (having some of one and some of another is causing test failures, wheres straight up swap seems fine)

PR #252, opened 2018-06-20
MAID-2688 fix/error: adds missing error handling for failed app_reconnect

hunterlester:


src/native/_nfs.js
```diff
@@ -53,22 +53,6 @@ const readFileInfo = (fileInfo) => {
...
-      }
-
-    } catch (e) {
-      // we can safely ignore this
-      if (console && console.warn) {
-        console.warn(`Parsing user metadata '${user_metadata_ptr}' of '${data_map_name}' failed: ${e}`)
-      }
-    }
-  }
-
```
hunterlester: Why remove this branch?
The addition of error handling at https://github.com/maidsafe/safe_app_nodejs/pull/252/files#diff-c2284895deccc698165325b396ad79b9R93, creates a decrease in code coverage however I'm not able to create a testing condition where `app_reconnect` throws an error in mock routing.
Thus my search for a way to increase coverage to make coveralls happy.
The above block of code is safe to remove because the condition can never be met since `user_metadata_ptr` will always be an instance of `Buffer`.
cc: @bochaco

src/native/_nfs.js
```diff
@@ -53,22 +53,6 @@ const readFileInfo = (fileInfo) => {
         ? new Buffer(0)
         : new Buffer(ref.reinterpret(file.user_metadata_ptr, file.user_metadata_len));
 
-  if (user_metadata_ptr) {
```
hunterlester: Why remove this branch?
The addition of error handling at https://github.com/maidsafe/safe_app_nodejs/pull/252/files#diff-c2284895deccc698165325b396ad79b9R93, creates a decrease in code coverage however I'm not able to create a testing condition where `app_reconnect` throws an error in mock routing.
Thus my search for a way to increase coverage to make coveralls happy.
The above block of code is safe to remove because the condition can never be met since `user_metadata_ptr` will always be an instance of `Buffer`.
cc: @bochaco

PR #256, opened 2018-07-25
chore/package: upgrade safe_app lib to v0.8.0

bochaco:


bochaco:
I think the problem simply was that the client libs were not published on S3 for Windows, whenever they do it this PR should pass the tests as expected.

hunterlester:
@bochaco `safe_client_libs` win resources are up now. Good to merge.

PR #257, opened 2018-07-27
MAID-2877 fix/validation: Improves error handling for malformed appInfo object passed to initialiseApp

hunterlester:
Currently, no errors are thrown on `initialiseApp` if a info object is passed, but if name (for eg) is `undefined`, it does not throw a useful error (but genAuthUri will die in an obscure fashion).

We should throw more helpful errors for object malformation.

```
info :

{ id : 'somthing, scope : null, name : undefined, vendor : 'something else' }
```

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -136,6 +136,8 @@ describe('External API', () => {
       const wrongAppInfo = Object.assign({}, appInfo, { customExecPath: 'non-array-exec-path' });
       return should(initialiseApp(wrongAppInfo)).be.rejectedWith('Exec command must be an array of string arguments');
     });
+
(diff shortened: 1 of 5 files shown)


test/index.js
```diff
@@ -136,6 +136,8 @@ describe('External API', () => {
       const wrongAppInfo = Object.assign({}, appInfo, { customExecPath: 'non-array-exec-path' });
       return should(initialiseApp(wrongAppInfo)).be.rejectedWith('Exec command must be an array of string arguments');
     });
+
+    it('throws error for appInfo property', () => should(initialiseApp({ id: 'id1', vendor: 'vendor' })).be.rejected());
```
bochaco: Perhaps better to rename the test to 'throws errors for appInfo missing property' and use `rejectedWith`?

Issue #262, opened 2018-08-17
Merge the WebID & RDF PoC code into a dev branch


bochaco:
Create a "dev" branch and merge the PoC code for WebID and RDF APIs

PR #263, opened 2018-08-17
#262 First PoC code for WebID & RDF APIs

bochaco:


bochaco:
I created issue #264 to solve the Lint errors that are being reported and failing the CI.

PR #276, opened 2018-08-31
Change docs to reflect testing condition loginForTest and simulateNetworkDisconnect

b-zee:
`useMockByDefault` is based only on `NODE_ENV` being `testing`. Can be influenced by `forceUseMock` too.

### DIFF
--- a/src/api/auth.js
+++ b/src/api/auth.js
@@ -375,7 +375,7 @@ class AuthInterface {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bochaco:
@b-zee , let's change it to `test`.
`testing` also works but strictly speaking we just require that the string starts with `test`: https://github.com/maidsafe/safe_app_nodejs/blob/master/src/helpers.js#L30

b-zee:
Changed it!

PR #293, opened 2018-10-12
feat/constant: Constant to automatically obtain correct file version for updates

### LINKED ISSUE #291: Passing zero to `nfs.update(...)` multiple times without error.
https://forum.safedev.org/t/implementing-nfs-api-rename/2109/13

I'm able to pass `0` as the version when updating a file multiple times without error.
`nfs.update(fileName, file, 0, userMetadata)`
Is this expected behavior?

hunterlester:
Resolves #291

### DIFF
--- a/src/api/emulations/nfs.js
+++ b/src/api/emulations/nfs.js
@@ -259,12 +259,8 @@ class NFS {
     const fileContext = file;
     return lib.dir_update_file(this.mData.app.connection, this.mData.ref, fileName,
                            fileContext.ref.ref(), version)
-      .then(() => {
-        if (version === CONSTANTS.GET_NEXT_VERSION) {
-          fileContext.version = version + 1;
-        } else {
-          fileContext.version = version;
-        }
+      .then((newVersion) => {
+        fileContext.version = newVersion;
       })
       .then(() => fileContext);
   }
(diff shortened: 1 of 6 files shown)


src/api/emulations/nfs.js
```diff
@@ -257,7 +260,13 @@ class NFS {
     }
     return lib.dir_update_file(this.mData.app.connection, this.mData.ref, fileName,
                            file.ref.ref(), version)
-      .then(() => { file.version = version; })  // eslint-disable-line no-param-reassign
```
bochaco: @hunterlester from what I understood we don't have to do anything here (i.e. no change at all here except for the documentation as you did above) as it's already taken care by the safe_app lib when the version passed is 0

hunterlester: @bochaco [dir_update_file](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/nfs.rs#L108) doesn't return the file context, in comparison to [file_open](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/nfs.rs#L152).

src/api/emulations/nfs.js
```diff
@@ -257,7 +260,13 @@ class NFS {
...
+      .then(() => {
+        if (version === CONSTANTS.GET_NEXT_VERSION) {
+          file.version = version + 1; // eslint-disable-line no-param-reassign
```
joshuef: If we're gonna have the rule: `eslint-disable-line no-param-reassign`, let's use it.

@hunterlester, you can avoid linting issues here by just creating a new variable at the top of the relevant block:

`let versionToSet = version // or perhaps 'newVersion' ?`

It'll also read nicer (IMO) for setting the version.

hunterlester:
ref: https://github.com/maidsafe/safe_client_libs/pull/676

src/api/emulations/nfs.js
bochaco: Ah ok @hunterlester , I was misunderstanding this. 
There is one thing which really concerns me here though, you are assuming the next version is the current one plus 1, which is not necessarily true, so it seems that we'll either have to fetch the new version from safe_app lib somehow, or propose to have the `dir_update_file` to return the file context?

PR #295, opened 2018-10-12
chore/deps_downloader: configuration to download both mock and prod libs without needing NODE_ENV=dev

### LINKED ISSUE #277: Install dev and prod libs by default
When `yarn`ing, you're a dev. 

Having to remember to add `NODE_ENV=dev` on install is a PITA.

I propose we download both by default and leave it up to the app to exclude the dir they don't want in their build...

hunterlester:
Resolves #277

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ before_install:
   - "[[ ${TRAVIS_OS_NAME} = linux ]] && export CXX=g++-4.8 || echo 'skipped'"
 
 install:
-- export NODE_ENV=dev
 - yarn
 script:
 - yarn lint
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -5,6 +5,9 @@
 - Add instructions to the README for how to generate API docs locally
 - Support for exposing experimental APIs when `--enable-experimental-apis` argument is passed, or when `enableExperimentalApis` flag is set in the initialisation option
 
+### Changed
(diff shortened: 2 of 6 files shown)


hunterlester:
Anyone wondering why this configuration works, it's because `NODE_ENV=development is by [default](https://github.com/maidsafe/deps_downloader/blob/master/cli.js#L30).

bochaco:
@hunterlester , this looks good to me, although after a bit of testing I have a proposal.

With this change you won't need to specify any env var to install the package, and the native libs for both mock and prod will be installed, that's cool. 
Then when running the app it can either use the `forceUseMock` flag in the initialisation options or use `NODE_ENV=test` to load the native lib which uses the mock vault. That's also cool.

But if you install the package with `NODE_ENV=dev`, what you get installed is only the native lib for live network, which means it's kinda not backward compatible. Thus, what if we changed the default in our `deps_downloader` to be `dev` instead of `development` (or both?) so this change is backward compatible for anyone who is still setting the `NODE_ENV=dev`?

hunterlester:
Tests won't run/pass until https://github.com/maidsafe/deps_downloader/pull/12 is merged.

bochaco:
ok, @hunterlester , we have the new deps_downloader publisbed at https://s3.eu-west-2.amazonaws.com/deps-downloader/deps_downloader-0.3.0.tgz

PR #300, opened 2018-11-07
feat/experimental: add the 'fetch' experimental API

### LINKED ISSUE #290: Expose `fetch` experimental function to fetch network data objects from a URI
Expose `fetch` experimental function using the feature implemented by issue #284, which can fetch any data object from the network and instead of returning an HTTP response data, return the JS object where the API can be used on, e.g. if the URI is pointing a MutableData it would then return a MutableData API object.

bochaco:
The `fetch` experimental API allows apps to fetch a network resource from a URL, specially needed by apps supporting linked data.
This function is enabled with the `--enable-experimental-apis` and it returns an object with:
- the target network resource (`content`)
- the type of the resource fetched (`resourceType`)
- and the parsed path from the provided URL (`parsedPath`)

Resolves #290

### DIFF
--- a/src/web_fetch.js
+++ b/src/web_fetch.js
@@ -172,6 +172,12 @@ const readContentFromFile = async (openedFile, defaultMimeType, opts) => {
   return response;
 };
 
+// Helper function which is able to fetch a resource from the network
+// using a URL. It parses the URL and calls the subName/publicName resolver helper
+// (it can also call other type of resolvers like XOR-URL resolver in the future),
+// returning the network object that it's found with the applied URL resolution,
+// along with the type of the resolved network object and the path that was
+// parsed out from the URL.
 async function fetchHelper(url) {
   if (!url) return Promise.reject(makeError(errConst.MISSING_URL.code, errConst.MISSING_URL.msg));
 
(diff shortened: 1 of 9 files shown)


hunterlester:
PR looks good, looking into why tests are failing.

bochaco:
Thanks @hunterlester for looking into that, the problem is in the safe_app v0.8.0 package in S3, we'll need to wait for that to be solved.

src/web_fetch.js
```diff
@@ -135,40 +150,70 @@ const readContentFromFile = async (openedFile, defaultMimeType, opts) => {
...
     endByte = (end === fileSize - 1) ? fileSize - 1 : end;
-    response.headers['Content-Length'] = lengthToRead;
-    response.headers['Content-Range'] = `bytes ${start}-${endByte}/${fileSize}`;
+    response.headers[HEADERS_CONTENT_LENGTH] = lengthToRead;
+    response.headers[HEADERS_CONTENT_RANGE] = `bytes ${start}-${endByte}/${fileSize}`;
   }
   return response;
 };
 
+async function fetchHelper(url) {
```
joshuef: @bochaco can you document this func a little as it's unclear from the naming what it's doing exactly.

bochaco: sure @joshuef , I'll push a commit with a comment for it

PR #312, opened 2018-11-26
chore/version: Version change to v0.10.0

bochaco:


b-zee:
Could the docs.maidsafe.net site also be updated to reflect the API?

bochaco:
Absolutely @b-zee , that'll be done today or tomorrow.

b-zee:
Cool!

Could I suggest that this the docs are automatically generated into GitHub Pages? For example [I've set this up in the library](https://github.com/project-decorum/decorum-lib/blob/994a57c9973f2997dcbe12c7070e673caf479f76/.travis.yml#L27) I'm working on: [project-decorum.github.io/decorum-lib/](https://project-decorum.github.io/decorum-lib/). Would be cool to see the documentation up-to-date with the latest tag automatically.

bochaco:
Yes, I think we can do something like that for sure, we have the script to do it already (`yarn docs`) we just need to hook the publishing side of it. Also, we would like to enhance the format of the html pages generated, I personally don't like it too much, so we may try to find what other tools are available for, if you know of any please let us know so we can take a look at it.

b-zee:
Have you tried the de facto standard [JSDoc](https://github.com/jsdoc3/jsdoc)?

I'm writing in TypeScript myself, so I can't recommend anything, except recommending to use TypeScript. ^^

theWebalyst:
I've not tried JSDocs, though somebody very quickly set that up for safenetworkjs, and then I stumbled on this which looks impressive:

https://github.com/pedronauck/docz

bochaco:
We are currently using JSdocs https://github.com/maidsafe/safe_app_nodejs/blob/master/package.json#L63, and I'm not an expert on it but the few times I spent some time understanding the options to document the APIs it doesn't offer much, and the webpages/htmls generated is not that great from my point of view, I don't know if I'm the only one who doesn't like it BTW

b-zee:
@bochaco It doesn't seem like you're using JSDoc?: https://github.com/maidsafe/safe_app_nodejs/blob/bdad4da5dc511c9d79394c42f6997d7ddb28643b/package.json#L11

You're using https://www.npmjs.com/package/documentation

bochaco:
oh yes, and this is the page that I remember looking at when trying to improve documentation: https://github.com/documentationjs/documentation/blob/HEAD/docs/RECIPES.md
So it seems they are kinda compatibles as per what the README says but not the same, in any case we can probably look at this in detail to see what's the difference with jsdocs and if that's better or just the same

theWebalyst:
JSDocs is set up for safenetworkjs, and I think the output is nicer than the SAFE API docs atm. I'm not sure, but expect this was the default setting /template. 

If you clone safenetworkjs and build the docs you can see for yourself.

PR #320, opened 2018-12-04
fix(Experiments): Web API exposure now throws per func

joshuef:
## Currently

The web API throws an experimental error (erroneously) when attempting to access `safeapp.web` on an instantiated safe app (when experiments are disabled).

## Wanted

safe_app should throw experimental errors when attempting to use exposed `web` functionality, (ie: `safeapp.web.getVocabs()` should throw an experimental error as opposed to attempting to access `safeapp.web`, which should not throw).

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -42,7 +42,10 @@ const isExperimentalApisEnabled = process.argv.includes('--enable-experimental-a
 // Helper function to check the experimental APIs flag and
 // either throw an error or log a warning message
 function checkExperimentalApisFlag(fn) {
-  const featureName = fn.name.replace('_', ' ');
+  const featureName = fn.name.replace('_', ' ')
+                        // bound added when binding context
+                        .replace('bound ', '');
+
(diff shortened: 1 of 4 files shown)


src/helpers.js
```diff
@@ -42,7 +42,10 @@ const isExperimentalApisEnabled = process.argv.includes('--enable-experimental-a
 // Helper function to check the experimental APIs flag and
 // either throw an error or log a warning message
 function checkExperimentalApisFlag(fn) {
-  const featureName = fn.name.replace('_', ' ');
+  const featureName = fn.name.replace('_', ' ')
+                        // bound added when binding context
+                        .replace('bound ', '');
```
bochaco: I don't see where 'bound ' is added, do we really need this `replace` statement?

joshuef: bound is seemingly added by the `bind(this)` added in app.js, such that the name becomes `bound getVocabs` (which is needed to provide the correct context without `call`.

PR #321, opened 2018-12-04
fix/fetch: when fetching a MD the `resourceType` value was 'NFS' instead of 'MD'

bochaco:
When calling the `fetch` experimental API to fetch a MD the value returned in the resourceType was 'NFS' instead of 'MD'. I'm also adding some tests for fetching different type of resources, i.e. MD, IMMD, RDF and NFS.

### DIFF
--- a/test/experimental_apis/fetch.js
+++ b/test/experimental_apis/fetch.js
@@ -14,6 +14,7 @@
 const should = require('should');
 const helpers = require('../helpers');
 const errConst = require('../../src/error_const');
+const { pubConsts: CONSTANTS } = require('../../src/consts');
 
 const createAuthenticatedTestApp = helpers.createAuthenticatedTestApp;
 const createUnregisteredTestApp = helpers.createUnregisteredTestApp;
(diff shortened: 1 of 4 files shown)


test/experimental_apis/fetch.js
```diff
@@ -54,11 +54,51 @@ describe('Experimental fetch function', () => {
...
 
-  // not supported yet, it needs support for XOR-URL
-  it.skip('fetch a MD resource', async () => {
+  it('fetch a MutableData resource', async () => {
+    const content = `hello world, on ${Math.round(Math.random() * 100000)}`;
+    const { serviceMd } = await createRandomDomain(content, '', '', app);
+    const info = await serviceMd.getNameAndTag();
+    should(info.xorUrl).not.be.undefined();
+    const networkResource = await unregisteredApp.fetch(info.xorUrl);
+    should(networkResource.resourceType).equal('MD');
```
joshuef: I think we `should(networkResource.content).equal( content )` or some such, to ensure we're grabbing the correct resource

test/experimental_apis/fetch.js
```diff
@@ -54,11 +54,51 @@ describe('Experimental fetch function', () => {
...
+      name: randomName,
+      nick: randomName,
+      website: `safe://mywebsite.${randomName}`,
+      image: `safe://mywebsite.${randomName}/images/myavatar`,
+    };
+    const webId = serviceMd.emulateAs('WebID');
+    await webId.create(profile);
+
+    const networkResource = await unregisteredApp.fetch(profile.uri);
+    should(networkResource.resourceType).equal('RDF');
```
joshuef: lets check the content here also

test/experimental_apis/fetch.js
bochaco: Let me see what I can do, but the content is the actual MutableData object, so perhaps I'll have to iterate thru the entries to see if they contain what were supposed to

test/experimental_apis/fetch.js
bochaco: Same as before, I'll give it a try to perhaps serialise it and compare, since the `content` is the RDF emulation object.

PR #325, opened 2018-12-06
refactor(docs): enhances API documentation

### LINKED ISSUE #313: Enhance API documentation
We need to document the experimental APIs with descriptive details and example snipets for how to use them. It could be worthy trying to also enhance the format of the webpage generated for the docs as discussed in PR #312.

### LINKED ISSUE #316: rdf.literal takes two arguments
`rdf.literal` should take a value as first argument and either a language specification or XSD datatype as second argment: https://github.com/linkeddata/rdflib.js/blob/master/src/data-factory.js#L30

Then can pass `rdf.literal( new Date("01/01/2037", "http://www.w3.org/2001/XMLSchema#dateTime")` and the correct `dataType` will be inferred.

### LINKED ISSUE #317: rdf.bnode should be rdf.blankNode
https://github.com/linkeddata/rdflib.js/blob/master/src/index.js#L49

### LINKED ISSUE #323: rdf.list is not a function
So far I don't see a `list` operation exposed in [rdflib.js](https://github.com/linkeddata/rdflib.js/blob/master/src/index.js)

### LINKED ISSUE #324: Adding provenance to graph store
Adding a `why` value to a statement to make a quad store, always results in the same NamedNode value: `chrome:theSession`.

hunterlester:
Resolves #324 
Resolves #323
Resolves #317 
Resolves #316 
Resolves #313

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "husky": "^0.14.3",
     "istanbul": "1.1.0-alpha.1",
     "jsdoc": "^3.5.5",
+    "jsdoc-escape-at": "^1.0.1",
     "mocha": "^3.2.0",
     "mocha-lcov-reporter": "^1.3.0",
     "rimraf": "^2.6.2",
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -13,7 +13,8 @@
         "includePattern": ".js$"
     },
     "plugins": [
-        "plugins/markdown" 
+        "plugins/markdown",
+        "node_modules/jsdoc-escape-at"
     ],
     "templates": {
         "referenceTitle": "SAFE Network Node.js Client Library",
--- a/src/api/emulations/rdf.js
+++ b/src/api/emulations/rdf.js
@@ -490,13 +490,13 @@ class RDF {
    *           const rdf = await mData.emulateAs('RDF');
    *           const id = rdf.sym("safe://pluto.astronomy");
    *           const serialised = `
-   *           @prefix : <#>.
-   *           @prefix ont: <http://dbpedia.org/ontology/>.
-   *           @prefix XML: <http://www.w3.org/2001/XMLSchema#>.
-   *           @prefix pro: <http://dbpedia.org/property/>.
-   *           @prefix res: <http://dbpedia.org/resource/>.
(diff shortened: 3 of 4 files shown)


joshuef:
@hunterlester is it possible to get the globals to appear at the top of the LHS bar?

I imagine `initialiseApp` being the msot looked for, but its sooo far down the bottom of the sidebar :(

bochaco:
Yeahh!! waaaay much better than the current one I'd say. What's your plan for the home page? it's empty now.

bochaco:
A very cosmetic issue, there is no margin at the very bottom of the LHS bar

bochaco:
In the global.html I see a section with all the type definitions, I'm not sure if we want this in the docs, if we do perhaps it should have its own page and a better format, these definitions are the errors the package handles internally and also those which it throws

hunterlester:
@bochaco I've added a commit that annotates methods with `@throws` to document the error type that it potentially throws. It's possible to add `typedefs` to the menu but I think it adds clutter and confusing to the docs. At key points in the docs I've linked to certain `typedefs` and errors so that devs can read about them.
Looking for a config that allows links to be to Github source code instead of generated HTML but don't see it yet.

@joshuef The current configuration doesn't let me move globals to the top of the menu but I've added an index page which should be a good starting point.

In general, the next thing I need to do is adjust styles. I don't thing that readily available in config options.

bochaco:
Sunds good @hunterlester . I also think we should double check about being able to override the order of the item in the menu, we are able to do so with current doc lib we use, otherwise we definitely need the index you say, and have an easy access to it from any page, I guess if the index is in the home page then fromany page you go to the home using the menu.

package.json
```diff
@@ -8,9 +8,7 @@
     "lint": "eslint src test",
     "fix-lint": "eslint --fix src test",
     "postinstall": "download_deps --package package.json",
-    "docs": "documentation build --config etc/documentation.yml --github true --output docs --format html src/**",
-    "serve-docs": "documentation serve --config etc/documentation.yml --github true --output docs --format html src/**",
-    "test": "cross-env NODE_ENV=test mocha --recursive",
```
bochaco: I don't think the `test` script should be removed

PR #326, opened 2018-12-07
fix/addPublicNameToDirectory: add publicName when the _publicNames has entries in old/non-RDF format

bochaco:
If the account has some entries in the `_publicNames` default container which are not RDF graphs, calling the experimental `addPublicNameToDirectory` function throws an error and the publicName is not added to the `_publicNames` container as expected.

### DIFF
--- a/src/api/web.js
+++ b/src/api/web.js
@@ -99,8 +99,6 @@ class WebInterface {
 
     const publicNamesContainer = await app.auth.getContainer(PUBLIC_NAMES_CONTAINER);
     const publicNamesRdf = publicNamesContainer.emulateAs('rdf');
-    const toDecrypt = true;
-    await publicNamesRdf.nowOrWhenFetched(null, toDecrypt);
 
     // Here we do basic container setup for RDF entries.
     // Doesn't matter if already existing, will just write same entries.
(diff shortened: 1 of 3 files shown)


test/experimental_apis/web.js
```diff
@@ -129,6 +129,19 @@ describe('Experimental Web API', () => {
       return should(publicNames).containDeep(['safe://_publicNames#thisIsATestDomain']);
     });
 
+    it('should add publicName even on top of container old data format', async () => {
```
joshuef: @bochaco from you explanation, is this test not reliant on there being _no_ webId data already created?

I'm not sure how we can know that there is no RDF there already from this test...

I think to be truly sure, we need a func to be clearing the `Mock Vault` in between tests, or?

src/api/web.js
```diff
@@ -247,21 +254,17 @@ class WebInterface {
...
       if (e.code !== errConst.MISSING_RDF_ID.code) {
         throw new Error({ code: e.code, message: e.message });
       }
     }
 
-    if (!existingRDF) {
-      directoryRDF.add(id, vocabs.DCTERMS('title'), directoryRDF.literal(`${PUBLIC_CONTAINER} default container`));
-      directoryRDF.add(id, vocabs.DCTERMS('description'), directoryRDF.literal('Container to keep track of public data for the account'));
-    }
+    directoryRDF.add(id, vocabs.DCTERMS('title'), directoryRDF.literal(`${PUBLIC_CONTAINER} default container`));
```
joshuef: There's no PUT penalty for re-adding this if already existing, correct?

test/experimental_apis/web.js
bochaco: Yes, probably, but I can jut simply create a new safeApp for that instance, no necessary to remove mock as the _publicNames container is per account, so with a new safeApp we know it's empty

src/api/web.js
bochaco: No, remember this is jut in memory, in the triple store handled in memory by rdflib.js, an from my research the tripple already exists then nothing happens. After we finished adding all triple to the in memory graph store, we serialise it for sending it to the MD, and that's where we effectively merge the graphs, i.e. in the RDF.commit function.

PR #328, opened 2018-12-12
fix/RDF: ignore any entry which cannot be decrypted when reading an emulated Private MD

bochaco:
When fetching the entries of the RDF emulated private MD, when an entry cannot be decrypted we are throwing an error but we can simply ignore such an entry and not use it to reconstruct the RDF graph/document, just log warning message.

Also, as a consecuence of the app revocation process, after revoking an app any entry it created in the `_publicNames` container has to be re-encrypted. The new encrypted entry's key string is different form the previous one (since a new encryption key is used), so the revocation process is removing the old one (soft-delete), and inserting the newly encrypted one. 
After re-authorising any other app, the new encryption key is obtained to decrypt the entries, but the old entry was soft-deleted and its key cannot be decrypted anymore as it was encrypted with previous encryption key. The new ones can be decrypted without issues by the other app/s after re-authorising themselves. We therefore should also simply not try to decrypt a key if the entry was soft-deleted (empty value).

This solves the issue seen in here: https://github.com/maidsafe/safe_browser/issues/470

### DIFF
--- a/test/experimental_apis/rdf.js
+++ b/test/experimental_apis/rdf.js
@@ -356,6 +356,7 @@ describe('Experimental RDF emulation', () => {
                                                 h.createRandomSecKey(),
                                                 h.createRandomNonce());
     const notMyRdf = notMyMd.emulateAs('rdf');
-    return should(notMyRdf.nowOrWhenFetched(null, encrypted)).be.rejectedWith('Core error: Symmetric decryption failed');
+    return should(notMyRdf.nowOrWhenFetched(null, encrypted))
+                                  .be.rejectedWith(errConst.MISSING_RDF_ID.msg);
   });
 });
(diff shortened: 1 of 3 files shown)


test/experimental_apis/rdf.js
```diff
@@ -356,6 +356,7 @@ describe('Experimental RDF emulation', () => {
                                                 h.createRandomSecKey(),
                                                 h.createRandomNonce());
     const notMyRdf = notMyMd.emulateAs('rdf');
-    return should(notMyRdf.nowOrWhenFetched(null, encrypted)).be.rejectedWith('Core error: Symmetric decryption failed');
+    return should(notMyRdf.nowOrWhenFetched(null, encrypted))
```
joshuef: @bochaco I'm unsure of this test.

The changes you made above, avoid decryption if there's no value, as keys were changed due to rejection of an app (as I understand).

Here you're simply attempting to decrypt with bad encryption keys... Which _should_ yield the previous error, and not 'no id', or?

PR #336, opened 2018-12-18
chore/docs: adds JSDoc template generated files

hunterlester:


hunterlester:
View docs here: https://hunterlester.github.io/safe_app_nodejs/

PR #341, opened 2019-01-10
chore/deps: upgrades FFI to master branch of repository

hunterlester:
The purpose of this is to be able to upgrade to Electron v4.0.0 in SAFE Browser

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -10,6 +10,7 @@
 ### Changed
 - All RDF methods which take RDF `subject`, `predicate`, `object` arguments now also take a fourth `provenance` argument, to be consistent with rdflib.js
 - Documentation enhanced with improved JSDoc annotation and template
+- Upgrade Node-FFI to depend on master branch of repository
 
 ### Added
 - Code examples for experimental API operations: RDF, WebId, WebInterface
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "cross-env": "5.1.3",
     "deps_downloader": "https://s3.eu-west-2.amazonaws.com/deps-downloader/deps_downloader-0.3.0.tgz",
     "enum": "^2.3.0",
-    "ffi": "^2.2.0",
+    "ffi": "https://github.com/node-ffi/node-ffi.git",
     "mime": "^2.0.3",
     "multihashes": "^0.4.14",
     "rdflib": "^0.19.1",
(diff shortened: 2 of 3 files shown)


joshuef:
I'm seeing some tests failing with node 10 on OSX. (though not many).

As this update facilitates 10, can you see how node 10 goers for you on windows, @hunterlester ? If we can have compat for node 8/9/10, that would be rad.

errors I'm seeing:

`  ✓ safe app lib contains "mock" dir if we force use of mock
Error: Dynamic Linking Error: dlopen(/home/mock/libsafe_app.dylib, 10): image not found`
An odd dir is being chosen (outwith of project folder)

and then (i'd wager related):
```
1) Smoke test is rejected if options object contains non-string configPath value:
     AssertionError: expected [Promise] to be rejected with a message matching 'Failed to set additional config search path. Reason: TypeError: error setting argument 0 - "string" must be a string, Buffer, or ArrayBuffer', but got 'Failed to set additional config search path. Reason: TypeError [ERR_INVALID_ARG_TYPE]: error setting argument 0 - The "string" argument must be one of type string, Buffer, or ArrayBuffer. Received type object'
      at next$onError (node_modules/should/cjs/should.js:1696:12)

  2) Encryption keys two app encryption throws error if decryptSealed not provided with cipher:
     AssertionError: expected [Promise] to be rejected with a message matching 'First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.', but got 'The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined'
      at next$onError (node_modules/should/cjs/should.js:1696:12)

  3) NFS emulation throws error if invalid user metadata type is passed while updating file:
     AssertionError: expected Function { name: 'test' } to throw exception with a message matching 'First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.', but got 'The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type object'
      at Assertion.fail (node_modules/should/cjs/should.js:258:17)
      at Assertion.value (node_modules/should/cjs/should.js:335:19)
      at Context.it (test/nfs.js:249:30)

  4) NFS emulation throws error if invalid user metadata type is passed while inserting file:
     AssertionError: expected Function { name: 'test' } to throw exception with a message matching '"value" argument must not be a number', but got 'The "value" argument must not be of type number. Received type number'
```

joshuef:
I've setup a PR https://github.com/hunterlester/safe_app_nodejs/pull/2 updating the test regex, which is failing as it differs between node versions.

PR #354, opened 2019-03-11
feat/XOR-URL: added `getXorUrl` experimental function to the ImmutableData Reader 

### LINKED ISSUE #353: Allow to get the XOR-URL of an ImmutableData fetched
Currently it's only possible to generate the XOR-UR of an immutable data only when it's being written to the network, more precisely when invoking the `close` function of the ImmDWritter. It should be possible and usefull to also be able to get the XOR-URL from an ImmDReader instance. This should be of course be allowed only when the experimental API flag is set.

bochaco:
Adding `getXorUrl` experimental function to the ImmutableData Reader which generates the XOR-URL of an already existing and fetched ImmutableData. Up until now you can get ImmD XOR-URL only when storing it on the network, with this you can get it after fetching it with its xorname.

Resolves #353

### DIFF
--- a/src/api/immutable.js
+++ b/src/api/immutable.js
@@ -100,8 +100,8 @@ class Reader extends helpers.NetworkObject {
    *     try {
    *         const cipherOpt = await app.cipherOpt.newPlainText();
    *         const iDataWriter = await app.immutableData.create()
-   *         const data = 'Most proteins are glycosylated.
-   *         Mass spectrometry methods are used for mapping glycoprotein.';
+   *         const data = `Most proteins are glycosylated.
+   *         Mass spectrometry methods are used for mapping glycoprotein.`;
    *         await iDataWriter.write(data);
    *         const iDataAddress = await iDataWriter.close(cipherOpt);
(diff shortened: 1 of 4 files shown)


src/api/immutable.js
```diff
@@ -79,6 +87,42 @@ class Reader extends helpers.NetworkObject {
...
+   * the XOR-URL as the codec of the content
+   * @returns {String}
+   * The XOR-URL of the ImmutableData.
+   * @example
+   * // Assumes {@link initialiseApp|SAFEApp} interface has been obtained
+   * const asyncFn = async () => {
+   *     try {
+   *         const cipherOpt = await app.cipherOpt.newPlainText();
+   *         const iDataWriter = await app.immutableData.create()
+   *         const data = 'Most proteins are glycosylated.
```
joshuef: For multiline text, I think ``` is cleaner that `'` (github is flagging this up as being an error, eg).

bochaco: let me take a look a this, and correct all docs that use`'` if it works with your suggestion

